PR #1, opened 2015-07-25
fixes stray = instead of == and adds linux case to terminal launching

ssfrr:


one-more-minute:
Awesome that you're playing around with this! Thanks for the patch.

PR #3, opened 2015-07-27
Add keybindings for linux/windows.

MichaelHatherly:
Mac-equivalent keybindings for Linux/Windows.

Have you decided on a keybinding for ``clear-inline-results``?

### DIFF
--- a/keymaps/julia-client.cson
+++ b/keymaps/julia-client.cson
@@ -22,3 +22,10 @@
  .platform-linux atom-text-editor[data-grammar="source julia"]:not([mini])':
   'ctrl-enter': 'julia-client:evaluate'
   'ctrl-shift-enter': 'julia-client:evaluate-all'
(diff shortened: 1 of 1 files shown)


one-more-minute:
I'm happy with these in theory, but have you checked they don't conflict with anything important? `Ctrl-.` will show you what keybindings are being activated.

MichaelHatherly:
``window:toggle-dev-tools`` is bound to ``ctrl-shift-i`` so we should probably pick something else.

one-more-minute:
`Ctrl-Alt-I` seems to be free on linux, so maybe we can just replace `Ctrl-Shift` with `Ctrl-Alt` on non-OS X.

MichaelHatherly:
Yeah, no conflicts there it seems.

one-more-minute:
Cool, thanks

PR #4, opened 2015-07-28
Add config for custom terminal.

MichaelHatherly:
This is a Linux-only config at the moment (I'd like it to be more general), wasn't sure how to set different defaults for each platform. I've set ``xterm`` as the default since I guessed it's probably installed in more places, happy to change it back to ``x-terminal-emulator`` though if that's the default you want.

Relatedly, have you though of leveraging [term](https://github.com/tjmehta/atom-term) or the [term2](https://github.com/webBoxio/atom-term2) fork instead of opening an external one? They seem to handle the Julia REPL fairly well.

### DIFF
--- a/lib/connection/terminal.coffee
+++ b/lib/connection/terminal.coffee
@@ -12,16 +12,14 @@ module.exports =
         console.log err
 
   term: (sh) ->
-    if process.platform == "darwin"
-      @exec "osascript -e 'tell application \"Terminal\" to activate'"
-      @exec "osascript -e 'tell application \"Terminal\" to do script \"#{@escape(sh)}\"'"
-    else if process.platform == "windows"
-      @exec "cmd /C start cmd /C #{@escape(sh)}"
-    else if process.platform == "linux"
-      # probably debian-specific
-      @exec "x-terminal-emulator -e \"#{@escape(sh)}\""
-    else
-      console.log 'unsupported platform'
+    switch process.platform
+      when "darwin"
+        @exec "osascript -e 'tell application \"Terminal\" to activate'"
+        @exec "osascript -e 'tell application \"Terminal\" to do script \"#{@escape(sh)}\"'"
+      else
+        @exec "#{@terminal()} \"#{@escape(sh)}\""
+
+  terminal: -> atom.config.get("julia-client.terminal")
 
   jlpath: () -> atom.config.get("julia-client.juliaPath")
   jlargs: () -> atom.config.get("julia-client.juliaArguments")
--- a/lib/julia-client.coffee
+++ b/lib/julia-client.coffee
@@ -8,6 +8,15 @@ utils = require './utils'
 completions = require './completions'
 frontend = require './frontend'
 
+defaultTerminal =
+  switch process.platform
+    when 'darwin'
+      'Terminal.app'
+    when 'linux'
+      'x-terminal-emulator -e'
+    else
+      'cmd /C start cmd /C'
+
 module.exports = JuliaClient =
   config:
     juliaPath:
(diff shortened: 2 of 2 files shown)


one-more-minute:
Great idea. How about setting the default to `'default'` and having `termpath` return something platform-specific if the config is unchanged. Then you can get rid of the windows clause from `term` as well, although we'll still have to special-case OS X for now.

@ssfrr What do you think of changing to `xterm`? We should probably go with whatever is more popular, although we can potentially do cleverer things like checking availability in future.

one-more-minute:
Also, the wording seems a little imperative for a config description – how about "Command used to open a terminal"?

ssfrr:
x-terminal-emulator is a [debianism] that symlinks "the system default terminal program", so I think that on debian/ubuntu systems it should probably be the safest bet.

[debianism]: https://wiki.debian.org/DebianAlternatives

MichaelHatherly:
> x-terminal-emulator is a debianism that symlinks

Thanks, should have just googled that. I'll leave that as is then. So long as customising is easy then defaulting to something common would be best.

> Also, the wording seems a little imperative for a config description – how about "Command used to open a terminal"?

That sounds better.

MichaelHatherly:
@one-more-minute the Windows ``cmd /C ...`` syntax is totally foreign to me so I'm unsure whether the change will actually work there. How would it handle spaces in a path, perhaps it needs escaping as well?

lib/julia-client.coffee
```diff
@@ -8,6 +8,15 @@ utils = require './utils'
...
 frontend = require './frontend'
 
+defaultTerminal =
+  switch process.platform
+    when 'darwin'
+      'Terminal.app'
+    when 'linux'
+      'x-terminal-emulator'
+    else
+      'cmd'
```
one-more-minute: Nice approach for this

lib/julia-client.coffee
```diff
@@ -22,6 +31,10 @@ module.exports = JuliaClient =
       type: 'boolean'
       default: true
       description: 'Enable notifications for evaluation'
+    terminal:
+      type: 'string'
+      default: defaultTerminal
+      description: 'Command used to open a terminal.'
```
one-more-minute: I would mention `(Windows/Linux only)`

lib/connection/terminal.coffee
```diff
@@ -12,16 +12,18 @@ module.exports =
...
-      # probably debian-specific
-      @exec "x-terminal-emulator -e \"#{@escape(sh)}\""
-    else
-      console.log 'unsupported platform'
+    switch process.platform
+      when "darwin"
+        @exec "osascript -e 'tell application \"Terminal\" to activate'"
+        @exec "osascript -e 'tell application \"Terminal\" to do script \"#{@escape(sh)}\"'"
+      when "windows"
+        @exec "cmd /C start #{@terminal()} /C #{@escape(sh)}"
```
one-more-minute: I was thinking of putting the whole `cmd /C start cmd /C` as part of the config. Probably also makes sense to include the flags, just in case alternatives need different ones.

MichaelHatherly: Sure, will do. Should be able to merge the two cases then I think.

one-more-minute:
Good work, thanks.

RE term2 – clearly it's a nice package, so it's good to know about. I imagine it wouldn't be too difficult to add a service to it and hook into that to run Julia with it. That said, right now I'm working on a rich console for Atom which will let us run a repl + have html results etc., so it might be worth waiting to see how that pans out and whether terminal integration is still a useful feature on its own.

MichaelHatherly:
> a rich console for Atom which will let us run a repl + have html results etc.

That would be even better. Just "emulating" a standard terminal seems like wasted effort if we can have something like that.

PR #5, opened 2015-08-05
DOC: note about apm install to get space-pen-views to build

spencerlyon2:
Atom was complaining that ink and julia-client couldn't find atom-space-pen-views.

Devs need to run `apm install` from repo dir to have node grab it


one-more-minute:
Ah yes, for some reason I always forget this step and it always bites me. Thanks!

PR #21, opened 2015-09-08
working interrupts on windows

Varanas:
This should enable interrupts on Windows that don't just kill the process. 
I'd be grateful if someone could test this (on Linux/OS X too, so that I know I didn't mess anything up with that part). 

Installation instructions can be found [here](https://github.com/JunoLab/atom-julia-client/tree/master/docs#developer-install), though only the `atom-julia-client` part is relevant.

@one-more-minute Sending the interrupt to the powershell-wrapper isn't ideal yet for a couple of reasons:

- It should be possible to jsut send it via node.js' internal functions without spawning another powershell instance, but I haven't figured that out yet.
- The port used for communication is hardcoded. Could use something like [this](https://gist.github.com/mikeal/1840641) to find a free port, though.


rohitvarkey:
@Varanas Just tested on OS X. Works fine!

one-more-minute:
This is a long-standing issue and I think lots of people will appreciate the improvement ;) As far as the first bullet goes I think it should be pretty easy to connect and send the signal with the [net functions](https://nodejs.org/api/net.html#net_net_connect_options_connectlistener), and it might make maintaining this stuff a bit easier. For the second I'm not too worried; we can fix that easily enough if it causes problems in future.

We might need to investigate the problems that happen when interrupting an idle client, but this is a strict improvement even without that fixed.

lib/connection/process.coffee
```diff
@@ -1,5 +1,7 @@
 process = require 'child_process'
 client = require './client'
+path = require 'path'
+net = require 'net'
```
one-more-minute: Looks like you don't use these imports?

Varanas:
I tried working with the `net`-functions (that's why they are imported), but something like 
```
client = net.connect(port: 26992, ->
            console.log('connected to server!')
            client.write("SIGINT"))
```
doesn't work. I get the `connected to server`, but my powershell script doesn't receive the SIGINT.

one-more-minute:
At a guess there's some [buffering](https://nodejs.org/api/net.html#net_socket_write_data_encoding_callback) going on. There doesn't seem to be an explicit flush method, but you could try `end()` or perhaps some incantation of `setNoDelay`.

Varanas:
Indeed. Thanks for the pointer, @one-more-minute, the implementation is a bit nicer now. 

@nilshg, could you try this new commit if you find the time? A bit of testing by someone else than me would be nice ;)

nilshg:
Will try to give it a go tomorrow if possible. I'm running a lot of hand-coded minimum distance estimators at the moment which often take hours to converge, and usually I find a stupid typo in the code after I've started running it, so interrupt would come in handy :)

Varanas:
Alright, I've just tested this on another Windows machine, squashed some bugs, refactored a bit and got rid of the debug `echo`s. I hope this doesn't break anything under Linux ;)

one-more-minute:
Sweet!

How about a news item to show off your work? ;)

Varanas:
Heh, why not ;)
Where'd that go? `NEWS.md`?

one-more-minute:
`Changelog.md` will do it

Varanas:
ah, I hadn't noticed that version 0.2 wasn't tagged yet

PR #22, opened 2015-09-08
typo

Varanas:
I think. Just stumbled upon that, but haven't seen any errors about it.


nilshg:
Maybe because there are no Win32 users around anymore?

Varanas:
Mh, but the comparison should be called for every OS, methinks. And iirc `Win32` is every Windows, not only x86.

one-more-minute:
Yeah, for some reason `process.platform == 'win32'` on all windows. Thanks for the fix!

PR #23, opened 2015-09-08
Update communication.md

EricForgy:
Link didn't work. I think this is the intended file.


one-more-minute:
Yup, just forgot to update that – thanks!

PR #26, opened 2015-09-13
RFC: Added doc page for workflow

spencerlyon2:
This is a first attempt at documenting how to use Atom to have a more productive workflow.

It is a bit rough and probably too verbose in places.

Comments welcome.


EricForgy:
Newbie question...

I was able to follow the instructions and get the Julia client working in Atom, so am interested in learning good workflows. But why does this documentation refer to Juno? Shouldn't it just refer to Atom?

Varanas:
The plan is to create a complete Julia IDE -- including `julia-client`, `language-julia` and so on -- which will then be called Juno, afaict.

Substituting "Atom" for "Juno"  wouldn't really be correct, since Juno introduces many features not found in Atom itself. Granted, it's a bit confusing that the LightTable based IDE has the same name, but that is only a transitory problem.

EricForgy:
Like I said, I am a newbie and am probably just confused, but I don't understand why installing a particular package would warrant a name change. What if the package is called "Juno" so I can just say, "I am using Atom with the Juno package installed." It feels a little strange to say, "I am using Juno" especially considering I will probably also use Atom for other language development.

If I get the story straight in my head, I can try to help with documentation.

Varanas:
I dunno if you've seen it, but there was a [package](https://github.com/one-more-minute/Juno-LT) for LT that not only provides the interaction with Julia, but a sensible set of default keybindings and some-other-stuff-I-can't-remember-anymore which differ from the defaults of the underlying text editor. 
That's not something I would expect from a package that "only" provides the ability to use another language in a nice way, but something more: It changes (well, is supposed to change) the program from "this is some text editor" to "this is a Julia IDE with extensive functionality".

So what I'm trying to say is this: The project is called Juno, this package here is just one part of it. Of course it still is just Atom with a bunch of packages installed and some settings and keybindings changed et ceterea et cetera, but it doesn't sound all that unreasonable to me to just give it a new name at that point.

docs/workflow.md
```diff
@@ -0,0 +1,149 @@
...
+type `Bar`. As you are hacking away on the fields and functionality of the `Bar`
+type you want to experiment with your work in the REPL. Here's what you might do:
+
+- Edit `foo.jl`
+- Fire up Julia and `include("foo.jl")` to load up your new types and methods
+- Return to the editor to make more changes to `foo.jl`
+- Go back to your REPL session and `include("foo.jl")` again, only to be greeted by the
+  following error message:
+
+```
```
Varanas: Maybe `3*[backtick]julia` instead, makes for better syntax highlighting.

docs/workflow.md
```diff
@@ -0,0 +1,149 @@
...
+ -->
+
+# Workflow
+
+One of the goals of this project is to help users adopt a more efficient
+workflow when editing Julia code. To understand how to gain this productivity
+boost, we will consider a common workflow for many Julia developers and a Juno
+based workflow.
+
+## Status quo
```
Varanas: While I kinda like the comparison to a plain-julia-workflow, I'm not entirely sure describing the status quo is necessary. It definitely is a nice read and now everyone knows why a Juno-based workflow is just plain superior ;)

docs/workflow.md
```diff
@@ -0,0 +1,149 @@
...
+the code from the current buffer directly into the chosen module. In our example
+above, this means that all the code we evaluate from `scratch.jl` is evaluated
+within the `Tmp` module _as if_ we had put the code in the module to begin with.
+
+### Pro Tips
+
+This workflow can take some getting used to, but can yield massive productivity
+boosts once you master it. Here are some tips to help as you learn:
+
+Don't ever set the current working module to the module defined in the buffer
```
Varanas: use a list for the tips too? Makes the formatting a bit nicer.

docs/workflow.md
spencerlyon2: I didn't do it here because I wasn't displaying Julia source code... just REPL output.

I'm happy to try it if you think it would be better.

docs/workflow.md
spencerlyon2: Agreed. I was a bit on the fence about it myself.

I think maybe we could drop it here and put in a line like "see the worfklow tips section of the manual for a description of the current workflow for many Julia developers." What do you think?

docs/workflow.md
spencerlyon2: I thought about that too. I'll implement that.

one-more-minute:
I think the current workflow section is a nice touch, actually – I wouldn't have thought to do it myself but I think it really drives home the motivation for Juno.

I also agree that calling it "the Atom workflow" wouldn't make much sense, any more than calling it "the Chromium workflow" would. Yes, that's the underlying base, but what we're building on top of it is largely orthogonal to it. The fact that you might use Atom for other languages is exactly the issue – if you're using haskell-ide, for example, you're still using Atom but you're definitely not using the workflow outlined here.

Anyway – really happy to see this, I think it will help people out a lot.

docs/workflow.md
```diff
@@ -0,0 +1,146 @@
...
+
+- Don't ever set the current working module to the module defined in the buffer
+you are working on.
+    - In the example above this means that if we wouldn't set the module to `Tmp`
+      from within `foo.jl`.
+    - Doing so would show `Tmp/Tmp` in the status bar and would make it awkward
+      to re-evaluate the entire buffer later on.
+    - If you do get in this situation, set the current working module to `Main`
+      to restore `Main/Tmp` in the status bar
+- Use the  `Julia Client: Work in file Folder` or `Julia Client: Work in project
```
one-more-minute: This tip should hopefully not be necessary after JunoLab/Atom.jl#6

spencerlyon2: Excellent. I'll delete this here once I see that PR has been merged

PR #27, opened 2015-09-14
Debounce module updates on cursor move

ssfrr:
This PR adds debouncing to cursor moves, so we don't try to update
the module we're evaluating into until the user stops moving the
cursor. I was noticing pretty high CPU usage and laggy performance
when typing or moving the cursor with the keyboard.

It also does some refactoring to modules.coffee. The `update` method
was being used to handle a bunch of different callbacks that did
different things depending on the state. This splits out into a
couple different callback handlers. `update` is now only for actually
sending the message to the julia process.

### DIFF
--- a/lib/modules.coffee
+++ b/lib/modules.coffee
@@ -5,17 +5,18 @@ module.exports =
   activate: ->
     @createStatusUI()
 
-    # configure all the events that might require setting or clearing the
-    # cursor move callback
-    atom.workspace.onDidChangeActivePaneItem => @activePaneChanged()
-    client.onConnected => @editorStateChanged()
-    client.onDisconnected => @editorStateChanged()
+    # configure all the events that persist until we're deactivated
+    @activeItemSubscription = atom.workspace.onDidChangeActivePaneItem => @activePaneChanged()
+    @connectedSubscription = client.onConnected => @editorStateChanged()
+    @disconnectedSubscription = client.onDisconnected => @editorStateChanged()
     @activePaneChanged()
 
   deactivate: ->
     @tile?.destroy()
     @tile = null
     @activeItemSubscription.dispose()
+    @connectedSubscription.dispose()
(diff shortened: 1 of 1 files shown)


lib/modules.coffee
```diff
@@ -5,17 +5,38 @@ module.exports =
...
     @createStatusUI()
 
-    @activeItemSubscription = atom.workspace.onDidChangeActivePaneItem =>
-      @update()
-    client.onConnected => @update()
-    client.onDisconnected => @update()
-    @update()
+    # configure all the events that might require setting or clearing the
+    # cursor move callback
+    atom.workspace.onDidChangeActivePaneItem => @activePaneChanged()
```
one-more-minute: You'll want to keep the `@activeItemSubscription =` part here so that it gets disposed of properly.

one-more-minute:
Looks like a decent improvement to me.

ssfrr:
I fixed the issue where I wasn't storing the `@activeItemSubscription`. I also now store the connected and disconnected subscriptions and clean those up too.

one-more-minute:
Ok, one last minor thing – could you clean up the subscriptions a little with [CompositeDisposable](https://atom.io/docs/api/v1.0.11/CompositeDisposable)?

(Technically I don't think it's as necessary to clean up the connection listeners, since the emitter will be destroyed as part of the julia-client cleanup, but it's no big deal either way)

ssfrr:
OK, done. I also realized that it wasn't unsubscribing from the other events on deactivation, so I added those as well.

I know just enough Javascript to be dangerous and am warming up to Coffeescript, but the Atom API is very foreign, so thanks for the pointers.

Re: the connection listeners - my thought was that this keeps things more symmetrical and doesn't rely on nonlocal information about how the client would behave.

one-more-minute:
Great, thanks for the very thorough work!

PR #34, opened 2015-09-17
RFC: First stab at module eval in console

spencerlyon2:
This is the atom side of having the console evaluate into the Julia module of the last open julia buffer.

see JunoLab/Atom.jl#8

### DIFF
--- a/lib/ui/console.coffee
+++ b/lib/ui/console.coffee
@@ -4,6 +4,7 @@
 {history} =     require '../misc'
 notifications = require './notifications'
 views =         require './views'
+modules =       require '../runtime/modules'
 
 module.exports =
   activate: (@client) ->
@@ -55,7 +56,8 @@ module.exports =
       @client.boot()
       @c.logInput()
       @c.done()
-      @client.rpc('evalrepl', code: editor.getText(), mode: mode?.name)
+      mod = modules.currentModule()
+      @client.rpc('evalrepl', code: editor.getText(), mode: mode?.name, module: mod)
         .then (result) =>
           @c.input()
           notifications.show "Evaluation Finished"


spencerlyon2:
ping! this has been working nicely for me

Varanas:
Whoops, completly forgot about this one, sorry.
While this certainly works, I'm not sure I'd really want this behaviour. It seems more convenient to me to either have the console always evaluate to `Main` or be able to choose the module to eval into (possibly with a setting to link the current editor's module with it). 
That said, I haven't done any work with this commit yet, so the above is purely theoretical ;)

one-more-minute:
Yeah, this is definitely a good step in the right direction but I'm not sure that this is the right interface either. It feels a bit too "spooky action at a distance" to me, e.g. if the module you're working in switches out from under you because you left the cursor in the wrong buffer (or in the wrong place within it).

How about something more similar to the way blank editors work – i.e. you start out in Main and can switch at will?

ssfrr:
I think that syncing the module of the editor and console makes sense,
but maybe if the console prompt changed to reflect which module you're
evaluating into it would feel more obvious what's going on?

spencerlyon2:
I'm with @ssfrr here. 

To me I use the secondary buffer as a more permanent scratchpad and then I move to the REPL if I want high-latency interaction. It has been really convenient to be able to seamlessly move from buffer to REPL and not have to change anything

spencerlyon2:
Thinking out loud here...

What if there was an option or something configurable in the settings that people could toggle on or off depending on whether or not they want the working module for the REPL to mirror the module from the last active Julia buffer? 

If people say yes, then my workflow described above would work seamlessly

If they say no, then we could implement what @one-more-minute suggests about having more state in the package that tracks the working module for the REPL (starts at Main, can be changed to anything just like Julia buffers).

Varanas:
I think having both would be great :)
So we can set the module manually and have some nice little icon next to it to automatically link the REPL's module to the current editor -- and since I can see use cases for both behaviours I wouldn't want to go into the settings menu each time I want to change it.

Varanas:
I think you can modify [this](https://github.com/JunoLab/atom-julia-client/blob/04ac902ed03db7295a49d0fcf36ae5e82672d1ba/lib/modules.coffee#L31-L43) (or something in that file... ;)) so that the module chooser shows up in the status bar at the bottom when the cursor is in the console, that'd be the best solution. If we were to throw a setting and into the mix you could then get this behavior as well.

spencerlyon2:
feature implemented

PR #39, opened 2015-09-19
Add a line on latex-completions to manual.

dpsanders:


Varanas:
Seems like a sensible change to me. Merging.

Issue #35, opened 2015-09-19
"Powershell not spawned"


dlfivefifty:
I get an error now (on master) saying "'powershell' could not be spawned.  Is it installed on your path", with a sidebar opening with error

Window load time: 1219ms
events.js:141 Uncaught Error: spawn powershell ENOENT

one-more-minute:
cc @Varanas

Varanas:
Yay, I broke it...
@dlfivefifty What OS are you on? I'd guess it's a Linux since `process.plattform` [seems to be problematic there](https://github.com/JunoLab/atom-julia-client/issues/17).
But I'm installing a Linux as I write, so hopefully I'll be able to debug this soonish.

SergeUK:
I am on windows 10.

Varanas:
And you have the same problem with spawning Powershell, @SergeUK?

SergeUK:
I don’t think so but to be honest with you I don’t really know how to try to spawn Powershell.

MichaelHatherly:
Ran across this as well this morning. Changing the order of the `process.platform` switches to check for `win32` first and default to `linux` seems to fix this on archlinux and windows 7 for me. 

```diff
diff --git a/lib/connection/process.coffee b/lib/connection/process.coffee
index 93d9cd5..049909e 100644
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -35,24 +35,24 @@ module.exports =
 
   spawnJulia: (port) ->
     switch process.platform
-      when 'darwin' || 'linux'
-        @proc = process.spawn(@jlpath(), [@jlargs()..., '-e', "import Atom; @sync Atom.connect(#{port})"])
-      else
+      when 'win32'
         @proc = process.spawn("powershell", ["-ExecutionPolicy", "bypass", "& \"#{__dirname}\\spawnInterruptibleJulia.ps1\" -port #{port} -jlpath \"#{@jlpath()}\" -jloptions \"#{@jlargs().join(' ')}\""])
+      else
+        @proc = process.spawn(@jlpath(), [@jlargs()..., '-e', "import Atom; @sync Atom.connect(#{port})"])
 
   interruptJulia: ->
     switch process.platform
-      when 'darwin' || 'linux'
-        @proc.kill('SIGINT')
-      else
+      when 'win32'
         @sendSignalToWrapper('SIGINT')
-
+      else
+        @proc.kill('SIGINT')
 
   killJulia: ->
[...]

PR #41, opened 2015-09-22
fix for `process` being shadowed by imports

Varanas:
causing `process.platform` to be undefined.
This leads to e.g. the default platform setting to be wrong on Windows.

ref #35


Varanas:
I'll just merge this now since it's only a bit of refactoring anyways.

PR #42, opened 2015-09-24
Keybindings for console require .ink-console css selector

spencerlyon2:
We were just using .conosle and that doesn't work anymore


ssfrr:
LGTM. Sorry about that. I made the change in the Ink package but didn't think to change it in atom-julia-client.

PR #52, opened 2015-10-09
Update workflow.md

EricForgy:
Minor cosmetic changes to intro paragraph. Replaced "buffer" with "file" to help newbies (like me), but left it "buffer" in the "Pro Tips" section.

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -1,14 +1,6 @@
-<!--
-TODO: should we call "this project" Juno? I saw in communication.md we did,
-but I'm not sure
- -->
-
 # Workflow
 
-One of the goals of this project is to help users adopt a more efficient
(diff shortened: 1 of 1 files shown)


EricForgy:
I might make some more changes, but wanted to gauge interest before spending time. One thing I might like to revise is the way to set the working module. Rather than suggesting to open just any file, I would suggest open a file something like runtests.jl borrowed from what I see and like in JuliaWeb. Then, you can simply click the module and choose your module rather than run a command.

spencerlyon2:
Looks good to me. Thanks

spencerlyon2:
RE: adding to main Julia docs.

I'm working on adding similar functionality (eval directly into module) as a REPL mode at the standard Julia REPL. When that happens I will definitely write up similar docs to live there also

PR #58, opened 2015-11-02
WIP: Connect to Julia server in a Docker container

malmaud:
Connect to a Julia process running in a Docker container, which can be local or remote. 

WIP since still needs error-handling logic.


spencerlyon2:
Is there a way to generalize this? Here's kinda what I have in mind:

- There is a simple command that starts a julia process like we have right now
- We can also start in docker containers like this PR allows
- There is an advanced command that pulls up some modal pane where users can enter server info, things like host, port, path to julia, other Julia args. Something like the [nuclide remote project dialog](https://github.com/facebooknuclideapm/nuclide-remote-projects/blob/master/images/ConnectDialog.png) is what I had in mind

Varanas:
This does (or rather should)  tie in with #8, right?

Some way to connect to a session with an arbitrary IP with a way to specify a custom start-up script (for starting docker or whatever) would be great, like @spencerlyon2 suggested.

Maybe you could try to generalize your approach, @malmaud?

PR #68, opened 2015-12-19
Add NoProfile option to PowerShell call

davidanthoff:
This will prevent any custom PowerShell startup scripts a user might have from running.

In my case I have code in my startup script that doesn't work when called from Atom, just generating errors that are seen in the julia console in Atom. This PR fixes that.


Varanas:
Thanks for this!

PR #69, opened 2015-12-26
Add commands for getting docs/methods

Varanas:
So I basically cleaned up and restructured what code I had for getting methods and documentation on a function or macro.

Atom.jl-side: https://github.com/JunoLab/Atom.jl/pull/17
atom-ink-side: https://github.com/JunoLab/atom-ink/pull/21

This design should be kinda okayish -- let me know what you think ;)

Also: We need default keybindings for those, I think. `Ctrl+D` and `Ctrl+M` are taken, so we could of course do something like `Ctrl+J Ctrl+D/M`. Suggestions?

### DIFF
--- a/lib/eval.coffee
+++ b/lib/eval.coffee
@@ -34,6 +34,31 @@ module.exports =
           @showError r, result.highlights
         notifications.show "Evaluation Finished"
 
+  # get documentation or methods for the current word
+  toggleMeta: (type) ->
+    editor = atom.workspace.getActiveTextEditor()
+    [word, range] = @getWord editor
+    # if we only find numbers or nothing, return prematurely
+    if word.length == 0 || !isNaN(word) then return
+    client.msg type, {code: word}, ({result}) =>
+      view = if result.type then result.view else result
+      view = @ink.tree.fromJson(view)[0]
+      @ink.links.linkify view
+      r = @ink?.results.toggleUnderline editor, range,
+        content: view
+        clas: 'julia'
+
+  # gets the word and its range in the `editor` which the last cursor is on
+  getWord: (editor) ->
+    cursor = editor.getLastCursor()
+    # The following line is kinda iffy: The regex may or may not be well chosen
+    # and it duplicates the efforts from atom-language-julia. It might be
+    # better to select the current word via finding the smallest <span> containing
+    # the cursor which also has `function` or `macro` as its class.
+    range = cursor.getCurrentWordBufferRange({wordRegex: /[\u00A0-\uFFFF\w_!´]*\.?@?[\u00A0-\uFFFF\w_!´]+/})
+    word = editor.getTextInBufferRange range
+    [word, range]
+
   evalAll: ->
     editor = atom.workspace.getActiveTextEditor()
     client.msg 'eval-all', {
--- a/keymaps/julia-client.cson
+++ b/keymaps/julia-client.cson
@@ -5,6 +5,7 @@
   'cmd-shift-j cmd-shift-m': 'julia-client:reset-working-module'
   'ctrl-`': 'julia-client:toggle-console'
   'ctrl-c': 'julia-client:interrupt-julia'
+  'cmd-j cmd-d': 'julia-client:toggle-documentation'
 
 '.platform-darwin .ink-console':
   'ctrl-c': 'julia-client:interrupt-julia'
@@ -26,6 +27,7 @@
   'ctrl-alt-j ctrl-alt-m': 'julia-client:reset-working-module'
   'ctrl-`': 'julia-client:toggle-console'
   'ctrl-shift-c': 'julia-client:interrupt-julia'
+  'ctrl-j ctrl-d': 'julia-client:toggle-documentation'
 
 '.platform-win32 .ink-console,
  .platform-linux .ink-console':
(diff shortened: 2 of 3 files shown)


Varanas:
Eh, `Ctrl-J Ctrl-M` is already taken. Not qiute sure what a nice default would be...

nilshg:
Suppose that would leave `Ctrl-J Ctrl-D`? Looks like a decent default to me.

Varanas:
Sure, but a shortcut for getting the methods would be nice as well.

nilshg:
Ah, didn't properly understand the question, probably a few too many Christmas drinks...
How about `Ctrl-J` `Ctrl-.` in a nod to dot notation for method calls?

lib/eval.coffee
```diff
@@ -34,6 +34,31 @@ module.exports =
...
+        clas: 'julia'
+
+  # gets the word and its range in the `editor` which the last cursor is on
+  getWord: (editor) ->
+    cursor = editor.getLastCursor()
+    # The following line is kinda iffy: The regex may or may not be well chosen
+    # and it duplicates the efforts from atom-language-julia. It might be
+    # better to select the current word via finding the smallest <span> containing
+    # the cursor which also has `function` or `macro` as its class.
+    range = cursor.getCurrentWordBufferRange({wordRegex: /@?[\w_!´]+/})
```
spencerlyon2: Hmm, I'm a little unsure about this one also. 

What about cases when I want to see the docs for something unexported from a module: e.g. `Base.mapreduce_pairwise_impl`(note this doesn't actually have docstrings, it's just something unexported)?

Varanas:
Fair point. I improved the regex a bit, it now does cover that case as well. 

Hopefully it doesn't break to much...

lib/eval.coffee
```diff
@@ -34,6 +34,31 @@ module.exports =
...
+        clas: 'julia'
+
+  # gets the word and its range in the `editor` which the last cursor is on
+  getWord: (editor) ->
+    cursor = editor.getLastCursor()
+    # The following line is kinda iffy: The regex may or may not be well chosen
+    # and it duplicates the efforts from atom-language-julia. It might be
+    # better to select the current word via finding the smallest <span> containing
+    # the cursor which also has `function` or `macro` as its class.
+    range = cursor.getCurrentWordBufferRange({wordRegex: /[\w_!´]*\.?@?[\w_!´]+/})
```
spencerlyon2: Will this regex work for greek characters in variable names?

I [noticed a while ago](https://github.com/willwhitney/hydrogen/issues/63#issuecomment-112490821) that the regex library used by plugins does not have `\w` match unicode; however, `\w` **does** match unicode in the grammar parsing (syntax highlighting) regex engine.

If this doesn't handle unicode, your hunch about finding `span` elements might be the way to go here.

Varanas: It doesn't. Sigh.

The `<span>` stuff won't work either because `atom-language-julia` doesn't recognise module name prefixes at all. So we'd need to either fix that there and twiddle a bit here or just fix the regex...

spencerlyon2: Hmm. I think it would be great to fix the grammar so it acts more like the parser, but that might be harder.

Given that you have a goal for what the scope selector should ideally do, do you think you could make the changes to the grammar?

Varanas: I'm at this regex now: `/[\u00A0-\uFFFF\w_!´]*\.?@?[\u00A0-\uFFFF\w_!´]+/` (per the link you gave me). It does seem to work, but I'm pretty sure there's lots of stuff in that char-range that isn't valid Julia.

I'll have a look at Julia's parser and see if there's an easy way of extracting valid charcodes...

As for changing the grammar: Is it possible to have nested scopes (haven't looked at it much so far)? Ideally we'd have something like 
```HTML
<span class="function"><span class="module">Base</span>.<span class="method">sin</span></span>
```
or similar so that it's easy to extract the whole function call.

Varanas: Hm, the regex I posted above seems quite nice...
regexing

I'm sure it has loads of problems, but it does seem better than anything else I could come up with. Extracting valid identifiers from Julia's parser is a bit above me, tbh.

spencerlyon2:
Nice! glad it seems to be working now.

I'm ok with this now, but would like a second reviewer opinion from someone else before merging

Varanas:
Yep, sure thing.

Thanks again for looking over all this! :)

one-more-minute:
I'm on board with this as a first pass, it's great to have something like this working again. In future it'd be good to look into repeating the Juno/LT approach; there are some functions in CodeTools.jl for pulling out qualified names and such which may be helpful.

Varanas:
Ah yes, for some reason I didn't even think about looking how you implented it in Juno/LT.

I think I'll just have a go at implementing this with `CodeTools.doc()` -- that'd save us the trouble with that funky regex as well.

Varanas:
Or not: `CodeTools.doc()` is a bit outdated and doesn't work on 0.4 anymore...

Are you okay with merging this now and improving upon it with `CodeTools` later on?

one-more-minute:
Yeah that's cool with me

Varanas:
Alright, then I'm gonna merge this now.

one-more-minute:
:+1:

PR #71, opened 2015-12-28
fix escaping bug on windows when opening REPL

Varanas:
`sh` would be escaped so something like `"\"julia\""`, which doesn't work. 

Dunno if this works on linux, someone might wanna check that ;)


hustf:
This is wonderful!

PR #77, opened 2016-01-06
Rejigger the config

one-more-minute:
Split into own file for organisational purposes, and adds options depending on the system so that we don't have any "only works on windows" options.

### DIFF
--- a/lib/config.coffee
+++ b/lib/config.coffee
@@ -27,7 +27,7 @@ if process.platform != 'darwin'
     order: 4
 
 if process.platform == 'win32'
-  config.spawnWrapper =
+  config.enablePowershellWrapper =
     type: 'boolean'
     default: true
     description: 'Use a powershell wrapper to spawn Julia. Necessary to enable interrupts.'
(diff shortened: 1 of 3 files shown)


Varanas:
Funny, I had the platform specific stuff implemented and not pushed yet ;) So, yeah, LGTM!

Btw, what does the notification option do?

one-more-minute:
If you're waiting for a long running task (evaluation, booting the client etc.) and switch to another window, that option just puts up a notification so that you can quickly switch back when it's done. It's quite nice but doesn't appear to work on windows for some reason.

Varanas:
That explains why I never noticed any difference between having it on and off... I'll try to remember looking into it sometime soonish.

one-more-minute:
[API reference](https://github.com/atom/electron/blob/master/docs/tutorial/desktop-environment-integration.md#windows). Notifications don't show up for me at all, and it sounds like they aren't easy to support on Windows versions below 10, so maybe we can switch to using the tray API.

PR #79, opened 2016-01-07
Remove JuliaParser instructions from README

davidanthoff:
The latest master of JuliaParser is now tagged, so the checkout should no longer be necessary.


ViralBShah:
This is nice!

PR #81, opened 2016-01-09
Support for bundled Julia

one-more-minute:
Putting together some bits so that we can dump Julia into the Atom app and have julia-client find it.

I'm also going to try and hack around the module precompilation issues so that we can bundle precompiled packages. Overall, I'm hoping that we can get the time to download Juno, install it and produce your first plot down to under a minute.

### DIFF
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -12,11 +12,14 @@ module.exports =
       path.join res, 'julia/bin/julia'
     if fs.existsSync p then p
 
+  packageDir: (s...) ->
+    path.join __dirname, '..', '..', s...
+
   initialiseClient: (f) ->
     packageDir = path.join __dirname, '..', '..'
     atom.config.unset('julia-client.initialiseClient')
     if (jlpath = @bundledExe())
-      proc = child_process.spawn jlpath, [path.join(packageDir, 'jl', 'caches.jl')]
+      proc = child_process.spawn jlpath, [@packageDir('jl', 'caches.jl')]
       proc.on 'exit', ->
         f()
     else
@@ -70,7 +73,7 @@ module.exports =
         return
       else
         cons.c.out "PowerShell version < 3 encountered. Running without wrapper (interrupts won't work)."
-    @proc = child_process.spawn(@jlpath(), ['-e', "import Atom; @sync Atom.connect(#{port})"], cwd: @workingDir())
+    @proc = child_process.spawn(@jlpath(), [@packageDir("jl", "boot.jl"), port], cwd: @workingDir())
 
   interruptJulia: ->
     if  process.platform == 'win32' && @useWrapper
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -12,6 +12,16 @@ module.exports =
       path.join res, 'julia/bin/julia'
     if fs.existsSync p then p
 
+  initialiseClient: (f) ->
+    packageDir = path.join __dirname, '..', '..'
+    atom.config.unset('julia-client.initialiseClient')
+    if (jlpath = @bundledExe())
+      proc = child_process.spawn jlpath, [path.join(packageDir, 'jl', 'caches.jl')]
+      proc.on 'exit', ->
+        f()
+    else
+      f()
+
   workingDir: ->
     paths = atom.workspace.project.getDirectories()
     if paths.length == 1
@@ -27,6 +37,8 @@ module.exports =
 
   start: (port, cons) ->
     return if @proc?
+    if atom.config.get 'julia-client.initialiseClient'
+      return @initialiseClient => @start port, cons
     client.booting()
     @spawnJulia(port, cons)
     @onStart()
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -30,8 +30,6 @@ module.exports =
       process.env.HOME || process.env.USERPROFILE
 
   jlpath: -> atom.config.get("julia-client.juliaPath")
-  # TODO: this is very naïve.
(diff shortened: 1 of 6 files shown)


Varanas:
Three things:
- I think it'd probably be better to use something like `atom.config.resourcePath` to navigate to the `resources` folder.
- The bundled julia should also be the default for all the other platforms, no?
- Minor naming nitpick: I'd rename `bundleExe` to `bundledExe` or something like that, then it doesn't sound like the function does or mutates something.

Generally, I think it'd be nice to check the existence of the specified executable no matter whether it is the default or not when starting julia. Then we can throw a nicer error if it fails.

one-more-minute:
* Makes sense.
* Of course – I should probably have made it clearer that this is a WIP ;) Once things are working on OS X I'll start testing on Windows.
* Also makes sense.

Definitely agree with checking the bundled executable, will put together a seperate PR with the `which` trick I mentioned in the gitter.

one-more-minute:
I've gone ahead and implemented the existence checking here as well. Would be good to have a quick set of eyes over this if possible, otherwise I'll merge later today!

lib/connection/process.coffee
```diff
@@ -1,8 +1,32 @@
...
+fs = require 'fs'
 
 module.exports =
+
+  bundledExe: ->
+    res = path.dirname(atom.config.resourcePath)
+    p = if process.platform is 'darwin'
+      path.join res, 'julia/bin/julia'
+    else if process.platform is 'win32'
+      path.join res, 'julia/bin/julia.exe'
```
Varanas: Do we only have OSX and Windows bundles? The linux case should be the same as OSX, no?

lib/connection/process.coffee
```diff
@@ -1,8 +1,32 @@
...
+    if fs.existsSync p then p
+
+  packageDir: (s...) ->
+    path.join __dirname, '..', '..', s...
+
+  initialiseClient: (f) ->
+    packageDir = path.join __dirname, '..', '..'
```
Varanas: This line is unnecessary, right?

lib/connection/process.coffee
```diff
@@ -11,28 +35,49 @@ module.exports =
...
+    which = if process.platform is 'win32' then 'where' else 'which'
+    proc = child_process.spawn which, [path]
+    proc.on 'exit', (status) ->
+      cb status == 0
+
+  jlNotFound: (path) ->
+    atom.notifications.addError "Julia could not be found.",
+      detail: """
+      We tried to launch Julia from:
+      #{path}
```
Varanas: Maybe mention that the path can be changed in the settings?

lib/connection/spawnInterruptibleJulia.ps1
```diff
@@ -1,11 +1,11 @@
...
-	[string] $jlpath, 
-	[string] $jloptions
+	[Int32] $port,
+	[string] $jlpath,
+	[string] $boot
 )
 
 # start Julia
-$proc = Start-Process $jlpath "$jloptions -e `"import Atom; @sync Atom.connect($port)`"" -NoNewWindow -PassThru
+$proc = Start-Process $jlpath @($boot, $port) -NoNewWindow -PassThru
```
Varanas: Is this valid?
Also, if all goes well, this file won't survive for long and be superseded be something actually working... :)

one-more-minute: Yup, works for me. Looking forward to something working!

lib/connection/process.coffee
```diff
@@ -11,28 +32,51 @@ module.exports =
...
-      text = data.toString()
-      if text then cons.c.err text
+
+    @checkExe @jlpath(), (exists) =>
+      if not exists
+        @jlNotFound @jlpath()
+        client.cancelBoot()
+        return
+
+      if atom.config.get 'julia-client.initialiseClient'
```
Varanas: Not sure I understand this setting -- is it only set in the bundles or on first start or something?

lib/connection/spawnInterruptibleJulia.ps1
Varanas: Yeah, I'll try hacking something together this evening -- Windows is a nightmare when it comes to doing something like that.

lib/connection/process.coffee
one-more-minute: Yes, I'll include the setting in the bundles and then it will only be active once, after the initial install.

Varanas: Makes sense.

Varanas:
LGTM then. Great work!

one-more-minute:
Thanks!

lib/connection/process.coffee
```diff
@@ -11,28 +32,51 @@ module.exports =
       process.env.HOME || process.env.USERPROFILE
 
   jlpath: -> atom.config.get("julia-client.juliaPath")
-  # TODO: this is very naïve.
-  jlargs: -> atom.config.get("julia-client.juliaArguments").split ' '
+
+  checkExe: (path, cb) ->
+    which = if process.platform is 'win32' then 'where' else 'which'
```
Varanas: @one-more-minute: This doesn't work on Windows when `path` is an actual path. I think we would need to call `where /r path` instead, but that doesn't work if `path` is only `julia`...

Once again, Windows is annoying. We should probably check for success of `spawn("$path", ["-h")` or something like that.

one-more-minute: We can just add an early return if `fs.fileExists(path)`. That will be far faster than shelling out on Windows anyway.

Varanas: Sure, that'd work as well.

Issue #76, opened 2016-01-06
Go To Definition.


mfcd:
Hi,

New to Juno and new to Atom.
Trying to get some sort of Go To Definition.

I have tried to configure the `goto` package but not sure how to proceed.
Due to performance issues, I'd like to avoid `Hydrogen`

Thanks.

spencerlyon2:
I use the [`atom-ctags`](https://atom.io/packages/atom-ctags) and have [this content](https://github.com/JuliaLang/julia/blob/master/contrib/ctags) my `~/.ctags` file.

There are some rough edges, but it does work pretty well

Varanas:
You can also use the `Toggle methods` command (at least on master), which will display all the method for any given function call with clickable links to their respective definitions.

We should probably use an interface similar to that of `atom-ctags` though, I like it -- doesn't require a mouse.

spencerlyon2:
Good point!

Could we have a command that gathers the info from `Toggle methods`, puts it in a modal (like you get from `cmd+shift+p` on osx) and lets the user select which method they want to visit?

Varanas:
Yeah, shouldn't be too much trouble to implement. I'll have a look.

Varanas:
Oh well, not as easy as I thought. I've made some progress, but haven't quite got this working.

In principle though this is sooo much nicer than my old implementation -- thanks for bringing it up!

Varanas:
Success:
![methodsview](https://cloud.githubusercontent.com/assets/6735977/12198592/c254b48a-b611-11e5-857a-5371eac149df.png)
Needs a bit of cleaning up and nicer display (something like the path or at least the file any given method's defined in), but basically works.

spencerlyon2:
Nice, that looks like a great start. 

I would also vote that you have the path

spencerlyon2:
This is the UX from `atom-ctags`.

<img width="452" alt="screen shot 2016-01-08 at 8 46 30 am" src="https://cloud.githubusercontent.com/assets/1001948/12201948/53791a56-b5e4-11e5-80e4-d7f23b0e7793.png">


I think that they do a good job with the line number and file - but I like the way you handled types. Maybe we could implement something similar for method location on top of what you have? Also can we drop the margins on the modal?

Varanas:
Sure, that's just style stuff and shouldn't be too hard.

[...]

PR #82, opened 2016-01-09
Much nicer method view panel

pfitzseb:
This implements a nicer and searchable frontend for the `toggle methods` command, as mentioned in #76.

methodview

[Julia side](https://github.com/JunoLab/Atom.jl/pull/20).
~~[Ink side](https://github.com/JunoLab/atom-ink/pull/25).~~

ToDo:
- [x] Default shortcut!
- [ ] style improvements, maybe? I'm open to suggestions. cc @spencerlyon2, @mfcd
- [x]  highlighting of matched keywords. That is actually not all that hard, see [here](https://github.com/atom/command-palette/blob/86741d106dbdc4860922e1b8bfded97dd699c274/lib/command-palette-view.coffee#L74-L92), but I didn't feel like copying their method just like that...
- [x]  Just thought of it -- this should probably be implemented with a `promise`, right?
- [x]  Errors should be handled gracefully instead of not at all.

### DIFF
--- a/keymaps/julia-client.cson
+++ b/keymaps/julia-client.cson
@@ -1,11 +1,12 @@
 '.platform-darwin .item-views > atom-text-editor[data-grammar="source julia"]':
   'cmd-enter': 'julia-client:evaluate'
   'cmd-shift-enter': 'julia-client:evaluate-all'
-  'cmd-j cmd-m': 'julia-client:set-working-module'
-  'cmd-shift-j cmd-shift-m': 'julia-client:reset-working-module'
+  'cmd-shift-j cmd-shift-m': 'julia-client:set-working-module'
+  'cmd-alt-j cmd-alt-m': 'julia-client:reset-working-module'
   'ctrl-`': 'julia-client:toggle-console'
   'ctrl-c': 'julia-client:interrupt-julia'
   'cmd-j cmd-d': 'julia-client:toggle-documentation'
+  'cmd-j cmd-m': 'julia-client:toggle-methods'
 
 '.platform-darwin .ink-console':
   'ctrl-c': 'julia-client:interrupt-julia'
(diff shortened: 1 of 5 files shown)


one-more-minute:
So to clarify, this is designed to replace the Juno/LT approach of just displaying the methods inline, right? Or is it more to augment it?

Varanas:
Yeah, pretty much. Could also be viewed as a replacement for `Ctrl-R` in plain Atom. 

This doesn't change the rendering of e. g.  `methods(sin) `.

Varanas:
Highlighting is implemented:
highlighted

If no one has any problems with the styling or any other suggestions, this should be good to merge.

Varanas:
We should really have a default shortcut for this!
Suggestion:
Use `C-Shift-J C-Shift-M` for selecting modules and `C-J C-M` for getting methods? Would be close to the `C-J C-D` we currently use for docs. Thoughts? cc @one-more-minute

spencerlyon2:
Quick question, will the path separator be `/` on unix systems?

Varanas:
Yup. Should be exactly the same as in backtraces and the like.

spencerlyon2:
OK great! I like your keymap suggestion

spencerlyon2:
What would it look like if each method had a shorter list item? It'd be nice to see more at a time

Varanas:
What about this?
unbenannt-1

spencerlyon2:
I like it

one-more-minute:
This is looking really good – big :+1: for screenshots ;) Any chance of having the sig on the left and the path on the right? There's also a method somewhere which will humanise package paths, i.e. `C:\Users\...\Lazy\...` :arrow_right: `Lazy\src\Lazy.jl` -- may well be worth applying that.

I've been thinking a bit how these things should fit together, so that all of our UI components have a clear purpose. Here's my proposal:

* We have one `docs` command which folds docstrings and methods together in a tree view (similar to the rendering of functions currently). Docstring is shown and methods are hidden by default for click-and-browse exploration.
* We also have a `jump to definition` command which loads up the appropriate file/location for the given object. If there's only one go straight there, if not launch this method view to navigate quickly to the right one. The command itself would no longer be method-specific.

How does that sound? It's not critical that this is all working thoroughly for this PR to be worth merging, of course, but I think that's what we should aim towards. (Open to other ideas of course!)

Varanas:
Yeah, that sounds like a good plan.

Couple of thoughts:
- This PR implements `methodswith(word)` when `word` is a type instead of a method -- would we drop that functionality or fold it into the docs command (not ideal, since that's not searchable)?
- How could `jump to definition` be implemented? Combination of the `symbols-view` approach (for local definitions) and `methods`?
- Do you have any ideas how refs ot other methods should work in the inline docs? Open a new pane or a browser window to display those or have it all inline? Matlab for example takes the first approach -- you get short `docs` with `help method` and clicking on the links to related methods opens a separate documentation browser. I think this may be better than doing all that inline, but I'm not sure.
- Imho getting to the definition isn't everything there is to a method view command -- it's also great for exploration of the available functions. We could conceivably have some way to open docs from there as well.
- The methodview logic should probably be used by autocomplete as well: It'd be super cool to get a list of methods (and maybe even a short doc string) after typing the opening paren, kinda like in VS:
vs
Of course we won't quite get to that level since Julia is a dynmic language, but I really like that direction.

Okay, went a bit on a tangent there, but a bit of brainstorming can't hurt ;)

Varanas:
This is now mostly superseded by https://github.com/JunoLab/atom-ink/pull/34.

PR #83, opened 2016-01-10
fix for not properly dealing with modules 

Varanas:
when requesting docs or methods for a given method.

This is on top of #82, because why not.


Varanas:
Incorporated into #82.

PR #85, opened 2016-01-12
WIP: Working interrupts on Windows

Varanas:
yay!

The c# code is horrendous though.

TODO:
- [ ] Error checking for input args as well as all functions! Potentially even a bit of `try catch` stuff.
- [ ] Refactor a bit, find a nice place for the exe.
- [ ] Properly redirect `STDOUT` and `STDERR`.
- [ ] Get rid of debug stuff.
- [ ] Test somewhere else.
- [ ] Figure something out for distributing this (like porting to powershell...)


Varanas:
Regarding distribution: At least [one other Atom package](https://github.com/yongkangchen/atom-ctags/tree/master/vendor) just ships binaries. So if I can wrap my head around compilation that shouldn't be much of a problem.

one-more-minute:
Nice! Will report back on whether it works properly on my system.

I think it simplifies the distribution question a lot if for now we just put the script itself in the repo and build it as part of the bundle build scripts – then the plugin can enable the wrapper if it exists in the resources folder. We can figure out how to make things work for people who install the plugins by hand later.

Varanas:
Yeah, I can also have a look at porting this to powershell -- calling the `SetConsoleCtrlHandler` works with 
```powershell
$MethodDefinition = @'
[DllImport("Kernel32.dll", CharSet = CharSet.Unicode)]
public static extern bool SetConsoleCtrlHandler(EventHandler handler, bool add);
'@
$Kernel32 = Add-Type -MemberDefinition $MethodDefinition -Name 'Kernel32' -Namespace 'Win32' -PassThru
$Kernel32::SetConsoleCtrlHandler($null, $true)
```
but I haven't yet tried writing the complete wrapper.

Varanas:
Now I feel silly. Got the powershell wrapper working...

PR #87, opened 2016-01-13
Modified ps script for working interrupts on Windows

Varanas:
and a couple of windows fixes in process.coffee.

@one-more-minute: The additional `cwd: @workingDir()` you added to the `powershell` invocation errors for me -- powershell cannot be spawned with it there.

I also added the early return in `checkExe` you suggested.

### DIFF
--- a/lib/config.coffee
+++ b/lib/config.coffee
@@ -26,7 +26,7 @@ if process.platform != 'darwin'
 if process.platform == 'win32'
   config.enablePowershellWrapper =
     type: 'boolean'
-    default: false
+    default: true
     description: 'Use a powershell wrapper to spawn Julia. Necessary to enable interrupts.'
     order: 2.5
 
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -107,9 +107,6 @@ module.exports =
       @proc.kill('SIGINT')
 
   killJulia: ->
-    if process.platform == 'win32' && @useWrapper
-      @sendSignalToWrapper('KILL')
-    else
       @proc.kill()
 
   sendSignalToWrapper: (signal) ->
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -4,7 +4,7 @@ net = require 'net'
 path = require 'path'
 fs = require 'fs'
 
-module.exports =
+module.exports = jlprocess =
 
   bundledExe: ->
     res = path.dirname(atom.config.resourcePath)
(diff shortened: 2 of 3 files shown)


one-more-minute:
Nice one. Is there any way for us to respect the working directory? Perhaps it could be passed through to powershell as an extra option?

Varanas:
Yeah, probably. I'll look into it (probably won't get to it until tomorrow though). 

Also, can you check that this works for you? Can't try it on another PC right now.

one-more-minute:
Yup, will let you know ASAP.

one-more-minute:
Works great!

Varanas:
Working dir should be set now.

There may or may not be a problem with `@workingDir()` though -- it seems like the current project folder can be a file. We should probably check for that.

one-more-minute:
One last issue: If the Julia session shuts down (e.g. `exit()`, the wrapper lives on, so Atom thinks Julia is still running. For now I've removed the indirection from the kill command to work around this. I'll see if there's a way to listen to process exits from ps.

Varanas:
The changes in the next to last commit wouldn't help with that, would they?
If Julia dies the wrapper still has no way of knowing that, but the change would keep Julia alive even though the wrapper is killed.

Varanas:
I thought something like 
```powershell
Register-ObjectEvent -InputObject $proc -EventName exited -Action {
	echo "Julia died, terminating the wrapper."
	Stop-Process $pid # Exit doesn't work either
}
```
might work, but it doesn't...

one-more-minute:
It occurs to me that we could also listen for client disconnects and kill the wrapper then.

Varanas:
Yup, those should indeed be handled in some way since they're just spamming the console anyways...

one-more-minute:
Let me know what you think of the latest commit and we can merge this.

Varanas:
LGTM, but haven't had the chance to try it yet.

Varanas:
It's been a long and arduous way, but we finally got there.

Varanas:
@one-more-minute Would you tag a patch release please? Actually living up to what we promise in the changelog seems to justify one ;)

one-more-minute:
I think the credit is all yours here, but yes, definitely glad to get there in the end. Will go ahead and tag a patch.

Varanas:
Thanks! I'll try to figure out what `apm` doesn't like about my setup or whatever so I don't have to bug you everytime when a patch should be tagged;)

PR #94, opened 2016-01-27
update syntax for block result support in ink

pfitzseb:
Depends upon https://github.com/JunoLab/atom-ink/pull/31 and Atom master.


pfitzseb:
Superseded by #144.

PR #95, opened 2016-01-28
randomly select a port for the wrapper communication on win

Varanas:


lib/connection/process.coffee
```diff
@@ -90,12 +90,19 @@ module.exports = jlprocess =
...
   spawnJulia: (port, cons) ->
     if process.platform is 'win32' and atom.config.get("julia-client.enablePowershellWrapper")
-      @useWrapper = parseInt(child_process.spawnSync("powershell", ["-NoProfile", "$PSVersionTable.PSVersion.Major"]).output[1].toString()) > 2
+      @useWrapper = parseInt(child_process.spawnSync("powershell",
+                                                    ["-NoProfile", "$PSVersionTable.PSVersion.Major"])
+                                          .output[1].toString()) > 2
       if @useWrapper
+        # get a random port between 1000 and 20_000; this may not be free, of course,
+        # but hopefully not very often
+        @wrapPort = Math.round(Math.random()*(20000-1000)+1000)
```
Varanas: It'd be great to do this in a smarter way, i.e. by actually checking if we have a free port. Couldn't figure out how to do that synchronously though, and any async calls will cause [this](https://github.com/JunoLab/atom-julia-client/blob/f52266ba7603c62c9aabd273556283fdd5169e58/lib/connection/process.coffee#L76-L89) to fail (or have `@useWrapper` undefined).

one-more-minute: How about passing a continuation to `@spawnJulia`?

one-more-minute:
This looks good, I'm going to go ahead and merge since it's clearly an improvement in itself, and we can always get the async stuff working later.

Varanas:
Continuations do work, but are ugly as hell. :) Oh well, this should be more correct than before, although I read somewhere that the port picked by `server.listen(0)` is guaranteed to be free on all OSs...

one-more-minute:
Whoops, forgot to actually do the merge. Anyway, this looks good, and I don't think the continuation is too bad.

Varanas:
Good thing you didn't, I probably wouldn't have tried to do this in a sensible manner otherwise ;)

EricForgy:
Silly question: The solution in getFreePort looks good, but is there a similar way to do the same thing in Julia? I'm considering some agent-based models and there could be a lot of ports used.

one-more-minute:
I think that if you don't specify a port, Julia will try to pick a free one for you. Could be wrong about that, but I've definitely seen some port-picking logic in Base somewhere.

Varanas:
[`listenany(123)`](http://docs.julialang.org/en/release-0.4/stdlib/io-network/#Base.listenany) should do the trick.

PR #99, opened 2016-02-02
move server.listen into continuation

### LINKED ISSUE #97: LoadError on Julia client startup/evaluation
Whenever I attempt to evaluate or set Julia-client to run on Atom's startup I get the following error in the console:

`ERROR: LoadError: connect: connection refused (ECONNREFUSED)
 in connect at /home/eghd/.julia/v0.4/Atom/src/comm.jl:23
while loading /home/eghd/.atom/packages/julia-client/script/boot.jl, in expression starting on line 421`

`Julia has stopped: 1, null`

I'm currently using Arch using Julia v0.4.3, both Atom.jl and CodeTools are on master, and the three Atom packages are up to date.

Varanas:
may or may not fix #97 (probably not, since I don't even know what causes it), but at least doesn't break anything. cc @one-more-minute

### DIFF
--- a/lib/connection/tcp.coffee
+++ b/lib/connection/tcp.coffee
@@ -36,6 +36,6 @@ module.exports =
       c.on 'data', @buffer (s) =>
         client.input JSON.parse s
 
-    @server.listen 0, 'localhost', =>
-      @port = @server.address().port
-      f?(@port)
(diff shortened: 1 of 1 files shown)


one-more-minute:
Looks like #97 is sorted?

Varanas:
Yeah.

PR #100, opened 2016-02-05
Module refactor + console eval

one-more-minute:
This is really much nicer. I could really go in for this whole model/view thing.

Eyes are appreciated!

### DIFF
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -1,10 +1,10 @@
+{Emitter} =     require 'atom'
 child_process = require 'child_process'
 net =           require 'net'
 path =          require 'path'
 fs =            require 'fs'
 
 client = require './client'
-{console: cons} = require '../ui'
 
 module.exports = jlprocess =
 
@@ -12,6 +12,7 @@ module.exports = jlprocess =
     @cmds = atom.commands.add 'atom-workspace',
       'julia-client:kill-julia': => @killJulia()
       'julia-client:interrupt-julia': => @interruptJulia()
+    @emitter = new Emitter
 
   deactivate: ->
     @cmds.dispose()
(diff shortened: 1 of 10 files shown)


spencerlyon2:
I haven't read the code super carefully, but module eval works great!

lib/connection/process.coffee
```diff
@@ -120,6 +120,9 @@ module.exports = jlprocess =
       server.close()
       fn()
 
+  onStdout: (f) -> @emitter.on 'stdout', f
+  onStderr: (f) -> @emitter.on 'stderr', f
```
Varanas: Not totally sure, but shouldn't these subscriptions be disposed of properly?

lib/runtime/modules.coffee
```diff
@@ -1,130 +1,153 @@
...
+    @dom = document.createElement 'span'
+    @dom.classList.add 'julia', 'inline-block'
+
+    @mainView = document.createElement 'a'
+    @dividerView = document.createElement 'span'
+    @subView = document.createElement 'span'
+
+    @dom.appendChild x for x in [@mainView, @dividerView, @subView]
+
+    @mainView.onclick = =>
```
Varanas: Could we change this so an `<a>` encompasses all `...View`s? Seems a bit nicer to me if the complete module description is clickable.

Varanas:
I like this a lot (implementation wise as well as the functionality)!

lib/runtime/modules.coffee
one-more-minute: The idea is that what's choosable == what's clickable. The main module is the bit that you actually choose, and it has no effect on the submodule since that's determined by the code in the buffer itself.

lib/connection/process.coffee
one-more-minute: In principle it'd be right to dispose of the listeners that are created [here](https://github.com/JunoLab/atom-julia-client/blob/omm/modules/lib/runtime/console.coffee#L32-L33). In practice it shouldn't matter, because when the package is deactivated the whole `Emitter` will be thrown away at once.

one-more-minute:
Thanks both for taking a look! If things look OK I can merge and start building on this tomorrow.

lib/runtime/modules.coffee
Varanas: Fair enough, I guess.

Varanas:
Go for it! :+1:

PR #101, opened 2016-02-07
ensure isinteractive() is true

jkroso:


Varanas:
This should also be changed [here](https://github.com/JunoLab/atom-julia-client/blob/master/script/spawnInterruptible.ps1#L15), I think.

jkroso:
I couldn't find any documentation on the `@()` syntax so I'm not sure I've done it correctly

one-more-minute:
Thanks for the patch, the syntax looks right. Is there are particular reason for adding this? I ask because we already set the flag [here](https://github.com/JunoLab/Atom.jl/blob/master/src/comm.jl#L18), but I'm not sure of the differences between the methods (we should probably choose on or the other).

One difference: the `-i` flag may cause more precompilation output when Atom.jl itself is compiled.

jkroso:
The only difference I was aware of was that this method enables you to use `isinteractive()` in your .juliarc. I wasn't aware of it changing the compiler's behaviour. Another solution might be to pass the "-f" option instead then load `.juliarc` after startup but before the first `eval`. That might be a problem for people who change the `LOAD_PATH` though

one-more-minute:
That makes sense. Lets go with this patch and remove the `eval` in Atom.jl. Could you rebase to sort out the merge conflict?

jkroso:
Didn't realise github doesn't do commit notification but just did another rebase

one-more-minute:
Ah great, yeah I probably should've kept an eye out – thanks!

PR #104, opened 2016-02-09
Implement block finding in Atom

one-more-minute:
This is a reasonable extra chunk of code but in the long run it will save us a lot; eventually I want to get rid of the ugly parsing stuff in CodeTools.jl completely as well as the dependency on JuliaParser.jl. Doing the parsing in the frontend has its own benefits, including a more responsive feel, less data to send over the wire, and a much simplified eval cycle. Eyes welcome!

### DIFF
--- a/lib/misc/blocks.coffee
+++ b/lib/misc/blocks.coffee
@@ -1,30 +1,8 @@
 # TODO: docstrings
 
+scopes = require './scopes'
+
 module.exports =
-
-  openers: ['if','while','for','begin','function','macro','module','baremodule','type','immutable','try','let']
-  reopeners: ['else','elseif','catch','finally']
-
-  @isKeywordScope: (ss) ->
-    for scope in ss.slice(1) # skip 'source.julia'
-      if scope.indexOf('keyword') > -1
-        return true
-
-  scopesForRange: (ed, range) ->
-    scopes = []
-    for l in ed.getGrammar().tokenizeLines ed.getTextInBufferRange range
-      for t in l
-        continue unless @isKeywordScope t.scopes
-        {value} = t
-        reopen = value in @reopeners
-        if value is 'end' or reopen
-          scopes.pop()
-        if reopen or value in @openers
-          scopes.push value
-    scopes
-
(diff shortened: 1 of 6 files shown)


lib/misc/blocks.coffee
```diff
@@ -0,0 +1,56 @@
+# TODO: docstrings
+
+scopes = require './scopes'
+
+module.exports =
+
+  getLine: (ed, l) ->
+    ed.getTextInBufferRange [[l, 0], [l, Infinity]]
```
Varanas: If this works reliably, we should probably replace all occurences of creating ranges that span a complete line with it instead of explicitly getting the line length.

one-more-minute: Good point – I only recently discovered this trick and there are definitely uglier versions of it lying around.

Varanas: [Yep](https://github.com/JunoLab/atom-ink/blob/master/lib/editor/result.coffee#L61). 

(I'll try to work through the rest of this soonish...)

one-more-minute: Ouch. Kill it with fire.

lib/misc/blocks.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+  isBlank: (l) -> l.match /^\s*(#.*)?$/
+  isEnd: (l) -> l.match /^end\b/
+  isCont: (l) -> l.match /^(else|elseif|catch|finally)\b/
+  isStart: (l) -> not (l.match(/^\s/) or @isBlank(l) or @isEnd(l) or @isCont(l))
```
Varanas: What about cases like
```julia
a(x) = 
x # cursor here, no indent
```
I mean, I know we can't reasonably do this without either relying on heuristics or calling `parse` on all lines, but we should probably note the cases we don't want to support somewhere...

one-more-minute: Cases like that seem so unidiomatic that it's probably ok not to support them. In the docs we'll just say that we support idiomatic code formats.

PR #115, opened 2016-02-21
add command 'julia-client:reset-workspace'

jkroso:
I found myself killing the process and running execute-all alot which is quite slow. This makes it pretty fast

### DIFF
--- a/lib/package/commands.coffee
+++ b/lib/package/commands.coffee
@@ -32,7 +32,10 @@ module.exports =
           juno.runtime.evaluation.toggleMeta 'methods'
       'julia-client:reset-workspace': =>
         if juno.connection.client.isActive()
-          juno.connection.client.rpc('clear-workspace')
+          editor = atom.workspace.getActiveTextEditor()
+          atom.commands.dispatch atom.views.getView(editor), 'inline-results:clear-all'
+          juno.connection.client.rpc('clear-workspace').then ->
+            juno.runtime.console.reset()
         else
           boot()
       'julia:select-block': =>
--- a/lib/package/commands.coffee
+++ b/lib/package/commands.coffee
@@ -30,6 +30,11 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.toggleMeta 'methods'
+      'julia-client:reset-workspace': =>
+        if juno.connection.client.isActive()
+          juno.connection.client.rpc('clear-workspace')
+        else
+          boot()
       'julia:select-block': =>
         juno.misc.blocks.select()
 
--- a/lib/package/commands.coffee
+++ b/lib/package/commands.coffee
@@ -31,13 +31,10 @@ module.exports =
           boot()
           juno.runtime.evaluation.toggleMeta 'methods'
       'julia-client:reset-workspace': =>
-        if juno.connection.client.isActive()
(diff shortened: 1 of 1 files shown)


lib/package/commands.coffee
```diff
@@ -30,6 +30,9 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.toggleMeta 'methods'
+      'julia-client:reset-workspace': =>
+        boot()
```
MikeInnes: Does it make sense to boot here? If the client has just been booted I would've expected calling `workspace()` to be redundant anyway – in which case `requireClient` makes more sense.

lib/runtime/evaluation.coffee
```diff
@@ -28,6 +26,12 @@ module.exports =
           @showError r, result.highlights
         notifications.show "Evaluation Finished"
 
+  evalCode: (text, line=1) ->
```
MikeInnes: Is there a particular reason for sending this information, or is this here just so that you can evaluate a code string? It might make more sense to implement a `workspace` command and call that via the `rpc()` – see for example the `exit` command.

MikeInnes:
Thanks for the patch! I'd like to be more clear on what specifically the issue is if possible.

Since there are keybindings for both `kill` and `eval-all` I would've expected that to be quite quick. I guess you're trying to replace this with the `workspace()` command to avoid Julia's boot time, and want there to be a command rather than typing it into the console?

jkroso:
yea thats right. For me it saved about 10s each restart on the project I was working on but I have a slow juliarc file so most people probably wouldn't gain that much. But it will always be faster since it allows Julia to reuse JITed code

jkroso:
I tried manually typing workspace() but I was doing it often enough that it got old real fast

lib/runtime/evaluation.coffee
jkroso: Yea it was just the most obvious way to do it at the time. I didn't understand how to do an rpc msg

lib/package/commands.coffee
```diff
@@ -30,6 +30,14 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.toggleMeta 'methods'
+      'julia-client:reset-workspace': =>
+        if juno.connection.client.isActive()
```
MikeInnes: Can replace this with `requireClient ->`

lib/package/commands.coffee
```diff
@@ -30,6 +30,14 @@ module.exports =
...
+          editor = atom.workspace.getActiveTextEditor()
+          atom.commands.dispatch atom.views.getView(editor), 'inline-results:clear-all'
```
MikeInnes: Not sure we should mix up resetting the workspace and clearing the console. This will error if you're not in an editor anyway, I think.

MikeInnes:
Sorry to nitpick, this is coming along nicely though!

lib/package/commands.coffee
jkroso: Yea if it just scrolled the console to the bottom that would be fine but not good deleting error messages

lib/package/commands.coffee
jkroso: Oh alright :)

PR #119, opened 2016-02-28
move to the last [row, col] in the next block

pfitzseb:
instead of preserving the column and staying on the first row. I realize the column-preserving probably happened for a reason, but it feels strange to me ;)

### DIFF
--- a/lib/misc/blocks.coffee
+++ b/lib/misc/blocks.coffee
@@ -54,11 +54,11 @@ module.exports =
       sel.setBufferRange [[last, Infinity], [last, Infinity]]
       sel.insertText '\n'
     # Move the cursor
-    to = row+1
-    {column} = sel.getHeadBufferPosition()
+    to = row + 1
     while to < ed.getLastBufferRow() and @isBlank @getLine ed, to
       to++
-    sel.setBufferRange [[to, column], [to, column]]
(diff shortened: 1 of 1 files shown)


MikeInnes:
Yeah I quite like the current behaviour, although I'm not adamant about it. Quick poll – anyone else have any preferences?

If you're just shooting down the file it makes no odds anyway – are there any situations where it would?

Varanas:
Well, I find it easier/faster to find out which code block will be executed -- when I'm reading the file top down, and follow the cursor movements, I always know up to where my code has been executed. Sure, indentation or even reading the block delimiters works, but isn't as fast.

Also, this change is something I made first thing after I first used the eval-and-move feature, so it seems like my expectations as to what it does weren't met. Dunno how this is for anyone else though.

pfitzseb:
Hm, polls don't work very well when there's only two people in them. :D
So yeah, bumping.

spencerlyon2:
If I had to choose I'd say move to the end

pfitzseb:
So, how about this? :)

MikeInnes:
Looks like I've been outvoted (though whether or not it's statistically significant... ;) Go ahead!

pfitzseb:
Well, twice as many people were in favour of this change... :D

Issue #124, opened 2016-03-03
Start up julia in the project's directory


tlnagy:
Right now, it seems as julia is always started in my home directory instead of the project directory so I have run `cd(/path/to/project/dir)` everytime I open Atom. Is there something misconfigured with my system or is this the expected behavior?

Varanas:
Depends -- do you only have one project dir open or multiple? If there's only one, we should automatically set the working dir, but if there are multiple we default to your HOME dir.

You can also use the `Julia Client: Work in File Folder` command or, alternatively, something like `cd(dirname(@__FILE__))`.

I'm not totally happy with that behaviour, but haven't invested much thought into a better heuristic yet --  do tell if you have any ideas! :)

tlnagy:
Actually, I think this is the latter case: I have multiple directories open. Wouldn't it make sense to default the julia kernel to the selected directory? Starting the kernel in the root directory doesn't seem useful to me.

pfitzseb:
So we'd set Julia's cwd to the directory containing the currently opened file? Hm, probably better than the current behaviour. I guess we could also try evaling every file's contents in it's on dir.

MikeInnes:
Defaulting to the first project folder if there are multiple seems like a good idea. As well as perhaps checking if the project folder is actually a file and jumping up a level.

pfitzseb:
Yep. (I still kinda think a project folder should never be a file...)

pfitzseb:
Alright, this has been implemented on master: We now start Julia in the first open project folder (or, if that is a file, in it's parent dir).

PR #127, opened 2016-03-08
make checking the julia executable more robust

pfitzseb:
and spawn julia in the first open project folder (or its parent if that is a file), as per [here](https://github.com/JunoLab/atom-julia-client/issues/124#issuecomment-193303789).

### DIFF
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -60,6 +60,7 @@ module.exports = jlprocess =
       # check whether path exists
       fs.stat p, (err, stats) =>
         if not err
+          # and is a file
           if stats.isFile() then resolve(); return
           # if it isn't, look for a file called `julia(.exe)`
           fs.readdir p, (err, files) =>
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -59,26 +59,18 @@ module.exports = jlprocess =
     new Promise (resolve, reject) =>
       # check whether path exists
       fs.stat p, (err, stats) =>
-        console.log err
         if not err
-          if stats.isFile()
-            resolve()
-          else
-            # if it isn't, look for a file called `julia(.exe)`
-            fs.readdir p, (err, files) =>
-              if err then reject()
-              if files.indexOf(@executable()) > 0
-                newpath = path.join p, @executable()
-                # check whether that file is no directory
-                fs.stat newpath, (err, stats) =>
-                  if stats.isFile()
-                    # change the `Julia Path` setting to that new path and carry on
-                    atom.config.set 'julia-client.juliaPath', newpath
-                    resolve()
-                  else
-                    reject()
-              else
-                reject()
+          if stats.isFile() then resolve(); return
+          # if it isn't, look for a file called `julia(.exe)`
+          fs.readdir p, (err, files) =>
+            if err or files.indexOf(@executable()) < 0 then reject(); return
+            newpath = path.join p, @executable()
+            # check whether that file is no directory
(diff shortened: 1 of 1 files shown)


pfitzseb:
@MikeInnes Actually, we might get away with deleting the fallback entirely -- I think this implementation here is quite comprehensive. What do you think?

lib/connection/process.coffee
```diff
@@ -36,28 +38,49 @@ module.exports = jlprocess =
...
+  # spawn Julia in the first open project folder (or its parent folder for files)
   workingDir: ->
     paths = atom.workspace.project.getDirectories()
-    if paths.length == 1 and fs.statSync(paths[0].path).isDirectory()
-      paths[0].path
+    if fs.statSync(paths[0].path).isFile()
+      path.dirname paths[0].path
     else
-      process.env.HOME || process.env.USERPROFILE
+      paths[0].path
```
MikeInnes: I think it's possible for their to be 0 project dirs, right?

lib/connection/process.coffee
```diff
@@ -36,28 +38,49 @@ module.exports = jlprocess =
...
     if p == '[bundle]' then p = @bundledExe()
     p
 
-  checkPath: (path) ->
-    new Promise (resolve) ->
-      fs.exists path, (exists) ->
+  checkPath: (p) ->
+    new Promise (resolve, reject) =>
+      # check whether path exists
+      fs.exists p, (exists) =>
```
MikeInnes: We might be able to flatten this big chain a bit / reduce the filesystem calls by using `fs.stat` to both check for existence and file-ness.

lib/connection/process.coffee
pfitzseb: True, didn't think of that. Should be fixed now.

lib/connection/process.coffee
```diff
@@ -73,12 +100,8 @@ module.exports = jlprocess =
...
     client.booting()
 
-    @checkPath(@jlpath()).then (exists) =>
-      if not exists
-        @jlNotFound @jlpath()
-        client.cancelBoot()
-        return
-
+    @checkPath(@jlpath())
+    .then =>
```
MikeInnes: We usually indent the `.then` and `.catch` to make it clear it's not a new statement

lib/connection/process.coffee
pfitzseb: Done. If we were to remove the `which` fallback it'd look even nicer ;)

lib/connection/process.coffee
pfitzseb: Also done.

MikeInnes:
Great changeset. I'm wondering about the exe-in-folder check though – is it really that common to input the folder just above the julia exe? It seems like there are a lot of small user mistakes like that that we could correct for, but at some point we have to give up and write better docs :)

pfitzseb:
Thanks for the review, @MikeInnes!

The exe-in-folder thingy came up at least twice, I think, so it seemed logical to fix it... Better docs would of course be nice, even though the setting explicitly says that we want the *binary*.

pfitzseb:
Alright, I've refactored stuff a bit -- if we want to keep the automatic checking for the exe then this should be good to go, otherwise I'll rip it out ;)

MikeInnes:
This looks good. I was vaguely worried that if we supported setting the path as a folder, users would rely on it and we'd never be able to change it back; but since you're actually resetting the path config to be correct that's actually not an issue.

Thanks for the patch!

PR #139, opened 2016-03-29
WIP: option to bypass atom's notification UI for errors

spencerlyon2:
This adds an option to have Julia runtime errors show up in the console rather than using Atom's notification UI when evaluating from the source file. 

It isn't finished as I didn't have a chance to hook up the part where we send the error to the console. I don't have time to dig into it myself right now, so if either @MikeInnes @pfitzseb  could fill in the `TODO` I logged to the console that would be great.

Thanks!

cc @simonbyrne

### DIFF
--- a/lib/connection.coffee
+++ b/lib/connection.coffee
@@ -9,11 +9,12 @@ module.exports =
     @client.boot = => @boot()
     @process.activate()
 
-    @client.handle 'error', (options) =>
-      if atom.config.get("julia-client.errorsToConsole")
-        @process.emitter.emit 'stderr', options.msg + '\n' + options.detail
-      else
-        atom.notifications.addError options.msg, options
+    atom.config.observe 'julia-client.errorsToConsole', (val) =>
+      @client.handle 'error', (options) =>
+        if val
+          @process.emitter.emit 'stderr', options.msg + '\n' + options.detail
+        else
+          atom.notifications.addError options.msg, options
 
     if atom.config.get("julia-client.launchOnStartup")
       atom.commands.dispatch atom.views.getView(atom.workspace),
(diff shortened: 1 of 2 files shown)


MikeInnes:
Looks good, and thanks for doing the skeleton of this even if it's not totally finished. I think the Simon was @SimonDanisch though :) Should be pretty simple to hook this stuff up.

That said, is this something we want to be choosable? As a rule I suspect it's better to pick the best option and stick with it. I think the right thing here might be to show errors in the console but have better notifications when the console has new content and isn't visible. Open to other ideas here though.

spencerlyon2:
Ahh, oops, My bad simons

I'm happy with defaulting to the console as I really don't like the big red boxes that I have to manually click with the mouse. That being said, we should come up with a way to notify users that the console has new output when it is not visible. Maybe a status bar icon with a subtle pulse or something?

MikeInnes:
Makes sense – I suggest having errors in the console as a power-user option for now and we'll make it the default (and remove the option) once we have better notification options for console output. A little icon in the status bar would do the trick, maybe it should turn red on errors and perhaps show a notification if the console isn't visible.

spencerlyon2:
That works for me. 

I'd be happy with that

pfitzseb:
Yeah, I like the status bar idea as well -- having an option seems fine in the meantime.

spencerlyon2:
@pfitzseb this didn't quite work:

```

pfitzseb:
@spencerlyon2 Yeah, my bad... forgot the `@`. Works now and prints the details as well.

On a general note: We should probably create some listeners for changed settings so they aren't only applied on startup/module load but whenever they are changed.

pfitzseb:
This should be good to go now -- I'll merge.

MikeInnes:
Thanks both!

PR #144, opened 2016-04-11
WIP: Docs

pfitzseb:
Preview: 
docs

I won't really have time to work on this for the next week, but thought I'd share what I have so far.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.5.0
+* Inline Documentation
+
 ## 0.4.0
 * Better Console buffering
 * History prefix support, like the repl
--- a/lib/misc.coffee
+++ b/lib/misc.coffee
@@ -2,6 +2,7 @@ module.exports =
   paths:   require './misc/paths'
   history: require './misc/history'
   blocks:  require './misc/blocks'
+  words:   require './misc/words'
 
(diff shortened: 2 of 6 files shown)


pfitzseb:
Alright, I've improved the code a bit in all parts of this changeset.

Also, we now have shiny marker ranges (which may or may not be too subtle) and the docs aren't rendered as a `Tree` anymore:
docs

I'd appreciate some comments about the look (and possibly feel, if you care to try it out) of this -- there are still some bugs, I think, and the code needs some polishing, but that should be done soonish.

pfitzseb:
Alright, this should be a bit better now. I'd appreciate you having a look, @MikeInnes.

(I'll resolve the merge conflict later, don't have any more time right now).

MikeInnes:
Yes, this looks good! (Sorry for the somewhat late review.) Will this work on stable Atom yet or should I go ahead and set up master?

That error behaviour sounds fine, let's run with that for now and change it if it causes confusion.

For branches I think we should just make 0.5 the master branch at this point, it's getting too messy otherwise. We can always make a julia-0.4 branch for back porting useful stuff.

pfitzseb:
This should work on stable, iirc. I still don't like the julia side of this, but it's better than nothing I guess...

PR #148, opened 2016-04-20
Julia As A Service

MikeInnes:
Cleans up the connection architecture and exposes the `client` object as a service.

What this means is that if packages want to extend what Juno can do – including by adding new ways of connecting to a client, e.g. over SSH – they can just grab the `client` object and get on with it. Equally, it might be possible to e.g. implement a profiler as an Atom plugin + Julia package on top of Atom.jl and julia-client.

Couple extras: better error messages, and a command which sends text to Julia's STDIN.

### DIFF
--- a/lib/connection.coffee
+++ b/lib/connection.coffee
@@ -23,7 +23,6 @@ module.exports =
         'julia-client:start-julia'
 
   deactivate: ->
-    @process.deactivate()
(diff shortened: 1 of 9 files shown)


ViralBShah:
Cc @mdpradeep @nkottary

Issue #151, opened 2016-04-25
Error in show displays as "Julia Client - Internal Error"


dlfivefifty:
Running the following in julia-client shows the error as a Julia Client - Internal Error, which makes it hard to debug:

``` julia
immutable Foo end

Base.show(io::IO,f::Foo)=error("My error")

Foo()
```

pfitzseb:
The "hard to debug"-part is fixed for now, but anything in user code should not be an internal error.

pfitzseb:
This should actually be fixed now.

PR #154, opened 2016-04-27
internal errors always rendered in console, notification is opt-out

pfitzseb:
motivated by #151.

Also change the recently introduced `Errors To Console` option to an opt-out setting for notifications for those kinds of errors.

julia side: https://github.com/JunoLab/Atom.jl/pull/34

cc: @dlfivefifty

### DIFF
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -50,12 +50,12 @@ module.exports =
       @callbacks[id].reject "cancelled by julia"
       @loading.done()
 
-    atom.config.observe 'julia-client.errorsToConsole', (toConsole) =>
-      @handle 'error', (options) =>
-        if toConsole
-          @stderr options.msg + '\n' + options.detail
-        else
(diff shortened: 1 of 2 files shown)


MikeInnes:
Ok, thinking about it, this change seems reasonable. But just as a note, we probably shouldn't focus too much on streamlining the Internal Error experience, as opposed to preventing them ;)

In this case we really need to catch display errors and treat them as user errors (I just keep putting it off).

pfitzseb:
Ye, sure, but nicer backtraces are valuable for us as well :)

Also, I'm not sure how or even if we can distinguish between internal and user errors in all possible cases.

MikeInnes:
Fair point. 

You can catch all user errors if you define them as "errors that occur during user-triggered eval/display". You might be thinking of e.g. a bug in Gadfly display as an internal error, but I just define us as users of our own system in that case.

pfitzseb:
I'll merge this then, unless you have any objections?

MikeInnes:
Please do

PR #164, opened 2016-05-06
get the correct versions packages

### LINKED ISSUE #159: pkgDir() may point to the wrong version's packages
[This](https://github.com/JunoLab/atom-julia-client/blob/82994b793c99ec7417017a0fe6f54996d9cc22c5/lib/misc/paths.coffee#L15-L20) always leads to the newest version's packages.

Crossref: http://discuss.junolab.org/t/open-package-in-new-window-doesnt-show-all-packages/581

pfitzseb:
This fixes #159. 
Only works with a functioning and correctly set julia executable, but I think that's a tradeoff everyone should be okay with. 

The regexp for getting the current version may or may not be a good idea -- no clue if Julias versioning scheme will change anytime soon. Maybe I should add a generic fallback similar to what we had before, but not sure. Opinions?

### DIFF
--- a/lib/connection/process.coffee
+++ b/lib/connection/process.coffee
@@ -196,6 +196,25 @@ module.exports =
         cwd: workingdir
       fn proc
 
+  getVersion: (fn) ->
+    p = new Promise (resolve, reject) =>
+      # not sure if the return here is necessary, but I think the rest of the
+      # code will be unnecessarily executed otherwise
+      if @version? then resolve(@version); return
+      @checkPath(@jlpath())
+        .then =>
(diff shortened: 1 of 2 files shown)


MikeInnes:
Yeah this seems like the right approach to me. Don't think it's likely that Julia's version scheme will change.

Only thing I would say is getVersion should probably be inside paths.coffee – we should try to avoid circular dependencies as much as possible.

pfitzseb:
Alright, done. Not totally happy with that placement, but we can change it later if we want anyways.

Thanks for having a look!

MikeInnes:
Thank you!

PR #168, opened 2016-05-10
limit the console history's size

### LINKED ISSUE #166: Console Wiping
From what I can tell the console keeps an infinite history unless you wipe it manually with `ctrl-j ctrl-c`. However, this is not robust because excessive printing can grind Juno/Atom to a halt because the memory used by the console can grow to huge numbers. This has happened to me many times, with common culprits being accidentally having too much debug printing (instantly knocks out Atom), and progress meter (since it does not write over itself, as noted in #165).

I think that the console should have a default size limit to handle this issue. It should be adjustable in the settings, and most of the time not be necessary, but is absolutely crucial for making the program more robust for long computations which on some regular basis print updates.

pfitzseb:
Fixes #166, requires https://github.com/JunoLab/atom-ink/pull/51.

Not sure how sensible that default is though...

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -47,13 +47,18 @@ config =
     default: proc.isBundled()
     description: 'Show Julia icons in the tool bar (requires restart).'
     order: 7
+  maximumConsoleSize:
+    type: 'integer'
+    description: "Limits the Console history's size."
+    default: 10000
+    order: 8
 
 if process.platform != 'darwin'
   config.terminal =
     type: 'string'
     default: terminal.defaultTerminal()
     description: 'Command used to open a terminal.'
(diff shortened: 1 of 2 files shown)


pfitzseb:
I'll go ahead and merge this soonish, if there are no objections? This should be strictly better than before.

MikeInnes:
Looks good! Especially since it's no longer naive :)

Does the limit need to be an option? Not a big deal either way.

pfitzseb:
Hehe, it's pretty much the same implementation I had before, but the console code has improved tremendously since then ;)

Having this be an option might be nice on slower machines, and as I said: I'm not sure if the limit I've chosen works well.

PR #181, opened 2016-05-24
install Ink via atom-package-deps

pfitzseb:
We could probably get rid of all those `withInk`s, but then again ink might be activated while Juno runs...


MikeInnes:
LGTM!

PR #183, opened 2016-05-27
WIP: rough draft of a notificator status bar element

pfitzseb:
Sneak preview:
notificator

Still a lot to do:
- [ ] Better code... ;)
- [ ] Nicer tooltip, obviously. Don't add repeated calls, but rather mark them with a `x N` or similar, better styling, etc...
- [ ] Figure out when we need notifications and when not (if console is openend and input is focused so the console scrolls, we don't really need any notifications).
- [ ] Figure out a nice way to automatically get short yet descriptive stack entries for the tooltip. Probably something like:
  - nothing for stdout and results that didn't error
  - first line of the error otherwise, truncated to `n` chars.
- [ ] Find a nice icon to replace the `Notif`.
- [ ] Improve the animation and stuff.
- [ ] Do we want to handle stuff here or in `Atom.jl`? A message handler will be needed anyways...

ink-stuff: https://github.com/JunoLab/atom-ink/tree/sp/notificator

cc @MikeInnes

### DIFF
--- a/lib/runtime/console.coffee
+++ b/lib/runtime/console.coffee
@@ -1,7 +1,7 @@
 {CompositeDisposable} = require 'atom'
 
 {history} = require '../misc'
-{notifications, views} = require '../ui'
+{notifications, views, notificator} = require '../ui'
 {client} = require '../connection'
 
 modules = require './modules'
@@ -22,6 +22,9 @@ module.exports =
         @c
 
     client.handle 'info', (msg) =>
+      notificator.notificator.notify
+        text: 'info'
+        type: 'msg'
       @c.info msg
 
     client.handle 'result', (result) =>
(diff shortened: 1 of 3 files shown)


MikeInnes:
I'm a little wary of adding more things to the status bar, but I guess I could see it working with just an icon.

See also https://github.com/suda/tool-bar/pull/143 for an alternative approach to notifying for the console – how does that look to you? (From the sounds of it we're going to have to implement our own toolbar, but that's basically doable.

pfitzseb:
Hm, I see where you're coming from, but with an icon it isn't all that much clutter:
notif

Having the notificator be the button for opening/toggling the console would obviously work as well, but since I'm no fan of the toolbar I'd like some alternative...

MikeInnes:
Fair enough.

I'm expecting that the toolbar might become more important as we need space for debugging buttons and stuff. What are your issues, and are they fixable?

pfitzseb:
I think I mostly find it unnecessary ;) That might of course change if there's some functionality in it that is most conveniently accessed in it, but that would probably only be status elements as opposed to commands (which can be accessed through the command pane anyways).

MikeInnes:
Going forward I see it being pretty useful for toggling debug mode and stuff. Like Chrome's devtools has a toggle for breaking on errors.

PR #184, opened 2016-05-31
Highlighting in results without external dependency

pfitzseb:
This seems like it may be general enough to have it in Ink, not sure though. 

I'd really like to get rid of the Highlights dependency soon, because it's necessary to patch master on Windows to get it running.


MikeInnes:
I agree, feel free to merge as you think it's ready. I'm guessing the duplicated code isn't going to change too often.

I've copied some of `views.coffee` over to Ink as a utility (because writing out `document.createElement, element.addClass` is horrible) but I'm not intending to expose it; I think it's probably out of Ink's scope to provide view rendering. If we wanted to generalise our view system we could always do it as an npm module.

PR #201, opened 2016-08-08
Cycler

MikeInnes:
Between Julia 0.4 and 0.5 boot time has gone up from ~5s to ~40s. This PR resolves this in several components.

`ipc.coffee` factors out our mechanism for remote procedure calls, which was previously tied to the Julia client. This is simply so that we can create an RPC communication channel with other kinds of server (and is used below).

`cycler.coffee` takes advantage of opportunities for concurrency between the user and the computer. If the user is working with an already-booted instance of Julia, we can use that time to boot up the *next* instance preemptively. This reduces boot time to sub-second. It gets even better; if the process finishes booting but still isn't needed, we can spend some time warming it up (by simulating the execution of common commands) and make it even more responsive.

`server.coffee` makes this even more ~~complicated~~ advanced by doing the above in an entirely separate process. This means startup time is saved even when opening fresh atom windows. It should also enable a much faster feedback loop when writing tests, which currently takes several minutes due to the need to boot multiple clients.

This has been working well for me and makes things feel much snappier and nicer, but given the extra surface area for issues we should probably roll it out gradually. Feedback and testing are of course appreciated.

### DIFF
--- a/lib/connection.coffee
+++ b/lib/connection.coffee
@@ -21,4 +21,4 @@ module.exports =
   boot: ->
     if not @client.isActive()
       @local.start()
-      time "Julia Boot", @client.rpc 'ping'
+      time "Julia Boot", @client.import('ping')()
--- a/lib/connection.coffee
+++ b/lib/connection.coffee
@@ -1,9 +1,10 @@
 {time} = require './misc'
 
 module.exports =
+  IPC:      require './connection/ipc'
   messages: require './connection/messages'
   client:   require './connection/client'
-  local:  require './connection/local'
+  local:    require './connection/local'
   terminal: require './connection/terminal'
 
   activate: ->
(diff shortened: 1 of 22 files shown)


pfitzseb:
I really like this change, even though the code got *slightly* more complex ;)

Apart from some spurious issue I had with interrupts not working (which I couldn't reproduce since), I only would change the cache size to something like 1 (maybe 2) -- I don't think it's hugely common to kill a julia process more than once every ~15s (or apparently ~40s on OSX), right?

MikeInnes:
Just a bit :)

It was around 40s for me on Windows as well. I often find needing a couple of clients in a row because I need to attempts to fix everything, but that might not be the common case. We can start with 1 and leave it as an option.

PR #211, opened 2016-08-16
require remote from electron

pfitzseb:
which required for this to work with newer electron versions (I think; will have to test more)

cc @ashnur, maybe you can try this branch sometime on Arch

### DIFF
--- a/lib/runtime/evaluation.coffee
+++ b/lib/runtime/evaluation.coffee
@@ -1,7 +1,5 @@
 path = require 'path'
-remote = require('electron').remote
-dialog = remote.require 'dialog'
-BrowserWindow = remote.require 'browser-window'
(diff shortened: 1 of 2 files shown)


MikeInnes:
Yeah this looks right to me, I know they changed the API at some point.

btw you can write this as `{remote} = require 'electron'`.

MikeInnes:
From the [docs](https://github.com/electron/electron/blob/master/docs/api/dialog.md) it looks like this should be:

```coffee
{dialog} = require('electron').remote
```

I guess we should also update any other remote requires for this change, too.

pfitzseb:
Right, was in a hurry to get that change in a branch... What I had there originally was pretty much the same as before -.-

PR #225, opened 2016-09-10
fix `pwd()` when used in `__init__()` or .juliarc.jl

jkroso:


pfitzseb:
Have you tried this on Windows? I think we had something similar to this some time ago (not sure why that got changed though) and iirc we had to pass the cwd into the powershell wrapper for this to work correctly.

jkroso:
No it's only tested in a Mac

MikeInnes:
Yes, this won't work on Windows as is. We changed away from this originally because it cause issues (I'm not sure specifically what it was).

An alternative solution to this would be to either (1) disable .juliarc.jl and load it after Juno has finished booting or (2) have a special .junorc.jl that does the same. I think I prefer (2) since it allows people to change package load paths and such. Feel free to have a go at implementing that or ask for any pointers.

jkroso:
Well you can detect Juno by testing if it's been loaded so the normal
juliarc will be fine.

I'll alter my PR to implement it that way

jkroso:
Ok its still only tested on OSX but I've made the changes

MikeInnes:
Looks solid but I'm still wary of preventing people from modifying load paths or other environment variables in a way that will affect Juno itself. How about loading juliarc as normal and loading `~/.junorc.jl` after Juno is loaded instead?

The include also needs to be in its own try/catch block so that user errors aren't classed as internal ones.

jkroso:
That does mean that when normal juliarc loads it will have the incorrect
pwd though. Which for me is a worse bug. Did passing the Cwd option to
node's spawn cause an error on windows or did it just effect the pwd?

jkroso:
OK so now I'm loading ~/.juliarc.jl with the cwd set by node's `spawn`. I know this doesn't work on windows but as long as it doesn't cause it to throw an error I see no harm in it. And setting the `pwd` in the boot script before loading a ~/.junorc.jl file if it exists. This should work on windows I'm guessing.

MikeInnes:
Ok, I think I misunderstood your original intent, I thought you were trying to change the pwd and wanted to see that reflected, but in fact you just want to see the pwd Juno is using reflected in the startup file. I think it amounts to the same in this case though.

Setting the cwd for the process doesn't work well when using the process server, because by its nature the process is spawned before we know what the cwd will be, so whatever you get is going to be meaningless. That's why we use a command after Julia has booted. One day I hope we can make the boot fast enough that the server isn't necessary, in which case we can remove the server and just pass the cwd as you want, but otherwise we have to stick to this approach.

I think that the way to solve this is to remove the cwd stuff, and to have a new `loadjunorc` command on the Julia side which is executed by the frontend after `cd`. That should give you a way to view the cwd without causing other issues.

jkroso:
It's used by my module loader, Kip.jl. But I might be better just to change
the way it works anyway

pfitzseb:
[](url)

Issue #220, opened 2016-08-29
Integration with IJulia kernel


habemus-papadum:
Hi -- back in the Juno/LT days, I used to have Juno/LT connect to a julia kernel that had been launched by IJulia (the flow involved running a command in Juno/LT that copied a small snippet  to the system clipboard, and then manually evaluating this text in a IJulia cell (or a manually launched repl session, or whatever suited your particular needs -- I happened to like IJulia))  This approach was often very productive (I can try to describe various pros and cons at some point if there is interest).  Looking at the v0.5 CHANGELOG, (and changes in Atom.jl) it looks like connecting to a julia process not started by Juno/Atom might be again possible -- I see references to "server" mode, etc.   Is this workflow possible again in Juno/Atom? (perhaps it was always possible?)  If not, any thoughts on how one might re-implement this behaviour?

thanks!

MikeInnes:
Yes, this is a nice feature, but one that got brushed aside a little with the recent changes. It shouldn't be too difficult to reimplement this though, so given the interest I may take a look at it soon.

lilinjn:
Ok -- thanks!  If there are beginner level steps that can be delegated, let me know.  Cheers

lilinjn:
Hi -- Using the new "Julia Client: Connect to Local Port" now makes the desired workflow possible.  Thanks!  

(Though the UI is not so easy to follow as of now -- for the time being here are my notes:
1) in the IJulia session/console repl that you want Juno to connect to:

````
import Atom
@async Atom.serve(8712) #use whatever port you like
````

2) (optional) use ssh to forward a local port to remote box if necessary; remember this port for the next step

3) In Juno -- "Julia Client: Connect to Local Port", make sure Julia Console is open -- type in the proper port number in the proper location next to the icon that looks like a clipboard.  

Upon success you should see a notification saying the connection to the local port was successful.

MikeInnes:
I'm implementing a much nicer version of this (like the old version) in https://github.com/JunoLab/atom-julia-client/pull/259#pullrequestreview-3351974.

lilinjn:
awesome!

PR #259, opened 2016-10-07
Connect to external processes

MikeInnes:
This adds a simple (but much nicer than before) UI for connecting from an external Julia process, e.g. in the terminal or IJulia.

### DIFF
--- a/lib/connection/messages.coffee
+++ b/lib/connection/messages.coffee
@@ -1,4 +1,5 @@
 client = require './client'
+tcp = require './process/tcp'
 
 module.exports =
   activate: ->
@@ -75,3 +76,19 @@ module.exports =
       else
         ""
       dismissable: true
+
+  connectExternal: ->
+    tcp.listen().then (port) ->
+      msg = atom.notifications.addInfo "Connect an external process",
+        detail: """
+        To connect a Julia process running in the terminal,
+        run the command:
+
+            using Juno
+            Juno.connect(#{port})
+        """
+        dismissable: true
+      client.onceAttached ->
+        if not msg.isDismissed()
+          msg.dismiss()
+          atom.notifications.addSuccess "Julia is connected."
--- a/lib/connection/messages.coffee
+++ b/lib/connection/messages.coffee
@@ -79,15 +79,16 @@ module.exports =
 
   connectExternal: ->
     tcp.listen().then (port) ->
+      code = "using Juno; Juno.connect(#{port})"
       msg = atom.notifications.addInfo "Connect an external process",
         detail: """
         To connect a Julia process running in the terminal,
         run the command:
-
-            using Juno
-            Juno.connect(#{port})
+        -
+            #{code}
         """
         dismissable: true
+        buttons: [{text: 'Copy', onDidClick: -> atom.clipboard.write code}]
       client.onceAttached ->
         if not msg.isDismissed()
           msg.dismiss()
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -61,6 +61,11 @@ module.exports =
   onAttached: (cb) -> @emitter.on 'attached', cb
   onDetached: (cb) -> @emitter.on 'detached', cb
 
+  onceAttached: (cb) ->
+    f = @onAttached (args...) ->
+      f.dispose()
+      cb.call this, args...
+
   isActive: -> @conn?
 
   attach: (@conn) ->
--- a/lib/connection/process/tcp.coffee
+++ b/lib/connection/process/tcp.coffee
@@ -16,9 +16,20 @@ module.exports =
       @listeners = @listeners.filter (x) -> x is conn
     conn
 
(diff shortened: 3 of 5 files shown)


lib/connection/messages.coffee
```diff
@@ -75,3 +76,19 @@ module.exports =
       else
         ""
       dismissable: true
+
+  connectExternal: ->
+    tcp.listen().then (port) ->
+      msg = atom.notifications.addInfo "Connect an external process",
+        detail: """
+        To connect a Julia process running in the terminal,
+        run the command:
```
pfitzseb: Maybe mention that the current process needs to be stopped first? Or re-add the `direquire` stuff below?

lib/connection/messages.coffee
```diff
@@ -75,3 +76,19 @@ module.exports =
...
+
+            using Juno
+            Juno.connect(#{port})
+        """
+        dismissable: true
```
pfitzseb: What would you think about adding a button that will copy the necessary commands to the clipboard? Shouldn't be too hard, see [here](https://atom.io/docs/api/v1.10.2/NotificationManager#instance-addInfo).

lib/connection/messages.coffee
MikeInnes: That's my oversight, fixed

lib/connection/messages.coffee
MikeInnes: Great idea, I've added this. It's something we can think about applying to our other buttons as well.

MikeInnes:
Thanks for the great feedback!

pfitzseb:
Anytime -- it's great to have this be so much more convenient!

Balinus:
Is there a how-to about how someine can connect to a remote julia session (through SSH)?

pfitzseb:
You can forward the local port used by Atom to the remote machine, see https://github.com/JunoLab/atom-julia-client/issues/220#issuecomment-250208410.

MikeInnes:
@Balinus If you need more instruction on this, feel free to open an issue on JunoDocs.jl and we can make time to document it in more detail.

Balinus:
Thanks, my first attempt didn't work. Here's what I did : 

1- Connect to the remote server through ssh with this command : "ssh -L localport:localhost:localport2 server.domain.com"
2- launch Julia on the remote server. Then launch command 
"import Atom"
"@async Atom.serve(localport2)" --> Got "juno-msg-readyTask (runnable) @0x00007fa0ccdce4a0"
3- on my local machine in Atom/Juno : "Julia Client: connect to local port" --> I'm asked to enter the port. I enter "localport"once I enter the port the julia local process is started but is not connected to the remote machine.

I will put this message in JunoDocs.jl. Thanks!

Both machines are on a LAN.

Balinus:
I've finally succeeded connecting to an external julia session on a remote machine. Here's my steps 

```
1. Launch Atom/Juno locally
2. Julia-client : connect to external process (Juno will then tell you which port to use locally --> we'll call it **PORT#2**)
3. Connect to remote server : "**ssh -R PORT#1:localhost:PORT#2 your.server.com**"
4. Launch julia in the terminal
5. type in julia session : "**using Juno; Juno.connect(PORT#1)**"
```
Hope this helps anyone trying to do that. It's a really nice feature (Atom on our server is sluggish and almost unusable).

Balinus:
Me again... :)

When Atom is connected to an external process, is there any way to define the backend gui for plots? Right now, I can't use any plotting functions.

```
julia> using Plots
WARNING: Method definition show(IO, Base.Multimedia.MIME{:text/plain}, Plots.Plot{#B<:Any}) in module Plots at /home/proy/.julia/v0.5/Plots/src/output.jl:170 overwritten in module Main at /home/proy/.julia/v0.5/Plots/src/output.jl:263.

julia> plot([1:10])
WARNING: No working GUI backend found for matplotlib.
/home/proy/.julia/v0.5/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/font_manager.py:1288: UserWarning: findfont: Font family [u'Helvetica'] not found. Falling back to Bitstream Vera Sans
  (prop.get_family(), self.defaultFamily[fontext]))
/home/proy/.julia/v0.5/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/figure.py:397: UserWarning: matplotlib is currently using a non-GUI backend, so cannot show the figure
  "matplotlib is currently using a non-GUI backend, "
```

AsafManela:
Thanks for the feature and how-to!

merl-dev:
> in Atom/Juno : Julia-client : connect to external process (Juno will then tell you which port to use locally --> we'll call it PORT#2)

Not sure what this refers to, since there is no such option through the menu, nor any such setting in the julia-client settings.

Balinus:
On my computer (Linux) I use the following shortcut for the command palette `ctrl-shift-p`. Then, you can write the command `external process`, which should give you the correct you.

PR #265, opened 2016-10-12
default keybinding for goto-symbol

pfitzseb:


pfitzseb:
@MikeInnes What do you think about those? The feature is kinda useful, so I'd like to have a sensible default keybinding for this.

MikeInnes:
Sure, I don't need this much myself but don't particularly object either.

PR #271, opened 2016-10-22
Progress refactoring take 2

pfitzseb:
Supersedes #226.

### DIFF
--- a/lib/ui.coffee
+++ b/lib/ui.coffee
@@ -1,10 +1,11 @@
 {CompositeDisposable} = require 'atom'
-{formatTimePeriod} = require './misc'
 
 module.exports =
   notifications: require './ui/notifications'
   selector:      require './ui/selector'
   views:         require './ui/views'
+  progress:      require './ui/progress'
+
 
   activate: (@client) ->
     @subs = new CompositeDisposable
@@ -16,25 +17,15 @@ module.exports =
     @subs.add @client.onDetached =>
       @ink?.Result.invalidateAll()
 
-    t_0 = {}
-
-    @client.handle 'progress!': (t, p) =>
-      if t is 'add'
-        t_0[p.id] = Date.now()
-      else if t is 'update' and not p.rightText?.length
-        p.rightText = formatTimePeriod (Date.now() - t_0[p.id])*(1/p.progress - 1)/1000
-      else if t is 'delete'
-        delete t_0[p.id]
-      @ink.progress[t] p
-
+    @client.handle 'progress!': (t, p) => @progress[t] p
 
   deactivate: ->
     @subs.dispose()
-    @ink.progress.emptyStack()
+    @progress.clear()
 
   consumeInk: (@ink) ->
     @views.ink = @ink
+    @progress.ink = @ink
 
-    indetProg = @ink.progress.indeterminateProgress 'indetJuno'
-    @subs.add @client.onWorking => @ink.progress.add indetProg
-    @subs.add @client.onDone    => @ink.progress.emptyStack()
+    @subs.add @client.onWorking => @progress.add progress: -1, id: 'indet'
+    @subs.add @client.onDone => @progress.clear()
--- a/lib/ui.coffee
+++ b/lib/ui.coffee
@@ -27,5 +27,8 @@ module.exports =
     @views.ink = @ink
     @progress.ink = @ink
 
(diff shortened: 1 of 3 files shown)


lib/ui/progress.coffee
```diff
@@ -0,0 +1,30 @@
...
+    @progs.push pr
+
+  update: (p) ->
+    pr = @progs.find (pr) => pr.id is p.id
+    pr.setProgress p.progress
+    pr.setLeftText p.leftText
+    if p.rightText?.length
+      pr.setRightText p.rightText
+    else
+      pr.setRightText formatTimePeriod (Date.now() - pr.t0)*(1/p.progress - 1)/1000
```
MikeInnes: It might be nice to formalise what happens in the right text more at some point. But this is the right approach for now, it lets people provide their own time estimates and stuff at least.

lib/ui/progress.coffee
```diff
@@ -0,0 +1,30 @@
...
+    pr.setMessage p.msg
+
+  delete: (p) ->
+    i = @progs.findIndex (pr) => pr.id is p.id
+    @progs[i].destroy()
```
MikeInnes: Need to handle the case where `p` is cleared already here.

lib/ui.coffee
```diff
@@ -15,14 +17,15 @@ module.exports =
...
+    @progress.clear()
 
   consumeInk: (@ink) ->
     @views.ink = @ink
+    @progress.ink = @ink
 
-    @subs.add @client.onWorking => @progress = @ink.progress.push()
-    @subs.add @client.onDone => @progress?.destroy()
+    @subs.add @client.onWorking => @progress.add progress: -1, id: 'indet'
+    @subs.add @client.onDone => @progress.clear()
```
MikeInnes: This will break things if you set up a background task via `@schedule for ...` (which is actually a pretty cool use case for this feature)

I would also put the progress-specific client handler and listeners in progress.coffee itself.

lib/ui.coffee
```diff
@@ -15,14 +17,15 @@ module.exports =
     @subs.add @client.onDetached =>
       @ink?.Result.invalidateAll()
 
-    @client.handle progress: (p) =>
-      @progress?.progress = p
+    @client.handle 'progress!': (t, p) => @progress[t] p
```
MikeInnes: The convention for these handlers is to use js-style naming since it works well in both languages

lib/ui/progress.coffee
```diff
@@ -0,0 +1,30 @@
...
+  progs: []
+
+  add: (p) ->
+    pr = @ink.progress.create p
+    pr.register()
+    pr.t0 = Date.now()
+    @progs.push pr
+
+  update: (p) ->
+    pr = @progs.find (pr) => pr.id is p.id
```
pfitzseb: You mean make `@progs` an object (`{}`)  and then do `@progs[p.id]` instead of the `find` stuff?

lib/ui.coffee
pfitzseb: Hm, not sure how to handle this then. We could of course only destroy the `indet` progress bar `onDone`, but that'll leave progress bars on the stack when e.g. interrupting a loop (which leads to progress bars not being destroyed).

lib/ui/progress.coffee
MikeInnes: Yeah that seems like it might be simpler

lib/ui.coffee
MikeInnes: Yeah that's an interesting case. Arguably just like closing files etc., progress bars should be wrapped in a `try`/`finally`

lib/ui/progress.coffee
```diff
@@ -0,0 +1,43 @@
...
+
+  progress: (p, prog) ->
+    pr = @progs[p.id]
+    return unless pr?
+    pr.setProgress prog
+    if pr.showTime then @rightText pr
+
+  message:  (p, m) -> @progs[p.id]?.setMessage  m
+
+  leftText: (p, m) -> @progs[p.id]?.setLeftText m
```
MikeInnes: I don't think that's necessary for now. If it actually becomes necessary to distinguish, we probably want something like grouping.

pfitzseb: Fair enough. Grouping sounds like a good idea, should there ever be some other language plugin that uses our porgress bars ;)

Issue #199, opened 2016-07-30
Feature request: improved display of output


marius311:
I'm fairly new to this plugin (although I really like it), so sorry if this has been discussed extensively elsewhere. 

I find it very difficult to read the output from evaluated lines in the editor. They're visually kind of in an awkward place (i.e. they all start at a different indentation). Long output requires me to move the mouse, click to expand, and often scroll (since they don't actually expand very far). 

Is there some way to improve the output? For example, I find what https://atom.io/packages/hydrogen is doing is really excellent. Could something like that be imitated?

pfitzseb:
Of course there are ways to improve the output, it's just that someone needs to implement them :)

If I understand you correctly, you'd rather have multiline results displayed below the current line? I guess that'd make Juno more notebook like, which would imo be better implemented along the lines of #63.

Then again, I agree that the results need to be improved a lot for better usability -- I'm just not quite sure how. [Here's](https://github.com/JunoLab/atom-ink/pull/59) a proof-of-concept PR for navigating the results with shortcuts though, so if we merge that and make the expansion a bit smarter so you actually see all you need, would that address (most of) your issues?

ChrisRackauckas:
Should this be closed due to the new fancy rendering of types done by Juno.jl?

marius311:
Sorry @pfitzseb I realize now I never responded to your question. Yes, more notebook-like would be exactly my suggestion. When you glance at a notebook, its super easy to distinguish what is input and what is output. In Juno, since each output starts at a different "x" position (at the end of the line), I find it very visually disharmonious. Just my 2 cents. 

A separate / related thing is, I don't understand why the output box is so small. E.g. in a notebook if I evaluate `rand(10,10)` I immediately see the entire thing in a output cell. In Juno, I have to 1) click the expand button then 2) scroll around b/c the output is so small. It can become quite a hassle. Why not just expand the output into rest of the window? That would still leave me having to click to expand the output (another reason why I like the "output cell" style), but at least it'd be better, and maybe that part wouldn't be so bad with the keyboard shortcuts you mentioned about. 

![image](https://cloud.githubusercontent.com/assets/1510968/18690655/7a46f4e4-7f8f-11e6-88fb-116c854a13e9.png)

[...]

PR #280, opened 2016-11-20
Adds option to display results as blocks

dustinrb:
Adds option in the settings menu to change whether results are displayed inline (current behavior) or underneath using the `type` option in `ink.Results`. See https://github.com/JunoLab/atom-ink/pull/93

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -51,6 +51,14 @@ config =
     description: "Limits the Console history's size."
     default: 10000
     order: 8
+  resultsDisplayMode:
+    type: 'string'
+    default: 'inline'
+    enum: [
+      {value:'inline', description:'Float results next to code'}
+      {value:'block', description:'Display results under code'}
+    ]
+    order: 10
 
 if process.platform != 'darwin'
   config.terminal =
--- a/lib/runtime/evaluation.coffee
+++ b/lib/runtime/evaluation.coffee
@@ -23,12 +23,12 @@ module.exports =
         [[start], [end]] = range
         @ink.highlight editor, start, end
         r = null
-        setTimeout (=> r ?= new @ink.Result editor, [start, end]), 0.1
+        setTimeout (=> r ?= new @ink.Result editor, [start, end], type: atom.config.get 'julia-client.resultsDisplayMode'), 0.1
         evaluate({text, line: line+1, mod, path: edpath})
           .then (result) =>
             error = result.type == 'error'
             view = if error then result.view else result
-            if not r? or r.isDestroyed then r = new @ink.Result editor, [start, end]
+            if not r? or r.isDestroyed then r = new @ink.Result editor, [start, end], type: atom.config.get 'julia-client.resultsDisplayMode'
             registerLazy = (id) ->
               r.onDidDestroy client.withCurrent -> clearLazy [id]
               editor.onDidDestroy client.withCurrent -> clearLazy id


pfitzseb:
What would you think about
```diff
 styles/ink.less | 46 +++++++++++++++++++---------------------------
 1 file changed, 19 insertions(+), 27 deletions(-)

diff --git a/styles/ink.less b/styles/ink.less
index 586600c..afaa33c 100644
--- a/styles/ink.less
+++ b/styles/ink.less
@@ -66,31 +66,8 @@ atom-text-editor::shadow {
   }
 }
 
-// TODO: more CSS sharing
-
-.ink.under {
-  font-family: @font-family;
-  color: @text-color;
-
-  cursor: default;
-
-  &.error {
-    color: @text-color-error;
-  }
-
-  background-color: @inset-panel-background-color;
-
-  border-top: 1px solid @syntax-wrap-guide-color;
-  border-bottom: 1px solid @syntax-wrap-guide-color;
-  padding-left: 5px;
-}
-
-
-.ink.inline {
+.result-mixin() {
   color: @syntax-text-color;
-  position: relative;
-  left: 1em;
-
   pointer-events: auto;
 
   &.error {
@@ -101,8 +78,8 @@ atom-text-editor::shadow {
   background: @syntax-background-color;
   border-left: 2px solid @background-color-info;
   border-right: 2px solid contrast(@syntax-background-color,
-                             darken(@syntax-background-color, 4%),
-                             lighten(@syntax-background-color, 4%));
+                                   darken(@syntax-background-color, 4%),
+                                   lighten(@syntax-background-color, 4%));
 
   border-radius: 3px;
 
@@ -111,7 +88,6 @@ atom-text-editor::shadow {
   opacity: 1;
   cursor: default;
 
-  max-width: 400px;
   overflow: hidden;
   white-space: pre;
 
@@ -170,6 +146,22 @@ atom-text-editor::shadow {
   }
 }
 
+.ink.inline {
+  .result-mixin;
+  position: relative;
+  left: 1em;
+  max-width: 400px;
+}
+
+.ink.under {
+  .result-mixin;
+  max-width: 100%;
+  white-space: pre-line;
+  padding: .5em;
+  border-top: 1px solid @syntax-wrap-guide-color;
+  border-bottom: 1px solid @syntax-wrap-guide-color;
+}
+
 .ink.stepper {
   .ink.inline();
   max-width: initial;
```

That way we get rid of all the duplication, and it looks kinda nice imo.

MikeInnes:
I'm fine with having this option. Any objections from anybody?

pfitzseb:
Nope, LGTM.

MikeInnes:
Thanks @dustinrb for the patch!

marius311:
Happened to glance at this repo again today and super happy to see this PR (it was me asking about something like this in https://github.com/JunoLab/atom-julia-client/issues/199). Going to try it out now, thanks.

pfitzseb:
You might also be interested in https://github.com/JunoLab/atom-julia-client/pull/283 then :)

marius311:
Have been using this today and I have to say, I absolutely love it. It makes the info inside the inputs os much more useful since its now easy to read / expand. Thanks very much! 

Here's some minor feedback I have since using it which is maybe of some use, but its really mostly minor. 

* The grey hashed highlighting of a line that occurs when you evaluate it is in the wrong place if any of the above lines have outputs. 

* Expanding an output can be kind of laggy, on my setup there's probably ~1 sec between clicking the expand button and when the lines below it shift down to make room. 

* If there is no output from a command, it might be useful to have the check mark smaller or suppressed or maybe even inline like before. It takes up a lot space and kind of breaks up the code (which I know was basically exactly the criticism I think you guys had before on implementing this at all :), and for the check-mark at least, I think you're right)

(This is Ubuntu Atom/1.12.3 with Ink/4c98938 and julia-client/2c5e412)

pfitzseb:
- Will look into that.
- I think this is an Atom core issue -- should hopefully get better with 1.13 afaik.
- What would you think about an "hybrid" mode which will display all single line and non-tree outputs inline and everything else as block results?

marius311:
For my money, I'd say keep even single-line outputs as blocks. So that basically if I know a command has outputs, my brain is telling me "look below to see the result" whereas if I know if doesn't, I don't want to look at anything. With your idea, I also need to process whether to expect my output to be single-line or not to know where to look. So hybrid mode with only check-mark's inline would be interesting to me. Although maybe I'm just being too picky at this point.

pfitzseb:
No, that makes sense.

Issue #63, opened 2015-11-05
Feature request: use special comment lines to divide code into notebook-like cells


cdsousa:
I love working in Jupyter but I miss a full fledged editor too many times.
On the other hand I'm also liking working in Atom+atom-julia-client but then I miss the custom block/cell code division.

I think the approach Matlab editor uses for solving this is a good trade-off.

Do you think it would be feasible to have a special comment line (e.g., `## this splits cells` or `#== this  splits cells ==#`) and a key shortcut to evaluate the cursor current cell?

Varanas:
:+1:
Not sure what syntax would be better, but I kinda like `##` since it's close to matlab and could be interpreted as markdown as well -- you jsut take a `h2` and structure your code with that.

cdsousa:
Let me add that displaying the results of each cell inline, right before the next cell division, would be awesome and would make Atom work more like a notebook. I guess that is harder from a technical perspective though.

spencerlyon2:
I like the idea of cells, good suggestion. 

But I also don't want my editor to become a notebook so I don't know that I'm on board with holding all results to the bottom of each cell. Given that we don't have a harsh of a visual break between cells I think it would be confusing to collect results at cell breaks

Varanas:
Yeah, I think I'd prefer the results to stay inline, i.e. have the cell-evaluation work like manually evaling all lines in the cell instead of working like the `eval file` command.

nilshg:
Maybe a compromise would be to have cell Eval when evaluating the cell division line, and line eval for all other lines?

spencerlyon2:
I think that another keybindings for cell eval is fine to distinguish
between line and cell eval. The discussion above was about where cell eval
output should go.

// Spencer

Varanas:
Once https://github.com/atom/atom/pull/9930 lands, we could easily implement @cdsousa's suggestion (along with a setting to switch to sequential inline evaluation of all lines in the cell).

CarloLucibello:
atom/atom#9930 has been merged

ChrisRackauckas:
Would this be a good time to revive this?

cdsousa:
I've just noticed that [Hydrogen](https://atom.io/packages/hydrogen) now supports this (just search for "cell" in the page).
It uses a syntax I'm not a big fan, but maybe if this gets implemented here it would be good to be compatible.

dgleich:
[...]

PR #283, opened 2016-11-26
Basic Cell-based Eval

JobJob:
Adds two commands: `run-cell` and `run-cell-and-move`. When either command is executed, the code in the current cell is eval'd. Results are displayed block style, below the cell delimiter.
Cells are delimited with `#---` by default, but this also adds an option to set the delimiter in the julia-client settings.

Gotcha: doesn't support anything to do with selections, probably won't do what users expect when the commands added are run when text is selected.

Credit to @dgleich for his [comment](https://github.com/JunoLab/atom-julia-client/issues/63#issuecomment-256341959) that really kick-started this.


dgleich:
Woohoo! Would love to see this merged. Let me know if there's anything I can do to help.

JobJob:
@dgleich you can use this right now if you like and see how it works for you:
```
cd ~/.atom/packages # or wherever your package dir is
cp julia-client path/to/somewhere/good/ #or "rm -rf julia-client" if you're feeling crazy
git clone https://github.com/JobJob/atom-julia-client.git julia-client
cd julia-client
git checkout job/basic-cells
```

Now you can restart Atom (or go to view->developer->Reload Window from the menu) and try it out. Use `#---` to separate cells (or change it in julia-client package preferences). Then you can set up keyboard shortcuts for `julia-client:run-cell` and `julia-client:run-cell-and-move` or run them from the command pallette with cmd-p or cmd-shift-p.

lib/misc/cells.coffee
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports =
+
+  getCurrentCell: (ed) ->
+    buffer = ed.getBuffer()
+    start = buffer.getFirstPosition()
+    end = buffer.getEndPosition()
+    regexString = atom.config.get 'julia-client.cellDelimiter'
+
+    unless regexString?
```
pfitzseb: `regexString` should never be `null` or `undefined`, no? `atom.config.get` could return an empty string, but that wouldn't be caught by this (and shouldn't matter anyways).

lib/misc/cells.coffee
```diff
@@ -0,0 +1,32 @@
...
+      return [start, end]
+
+    regex = new RegExp regexString
+    cursor = ed.getCursorBufferPosition()
```
pfitzseb: Should probably iterate over all cursors in the current editor, just like in `blocks.coffee`.

lib/misc/cells.coffee
```diff
@@ -0,0 +1,32 @@
...
+module.exports =
+
+  getCurrentCell: (ed) ->
+    buffer = ed.getBuffer()
+    start = buffer.getFirstPosition()
+    end = buffer.getEndPosition()
+    regexString = atom.config.get 'julia-client.cellDelimiter'
+
+    unless regexString?
+      return [start, end]
```
pfitzseb: This won't work -- you're not returning an object with `range` and so on as it's fields. Not that it matters, since this will never be called anyways.

lib/package/commands.coffee
```diff
@@ -22,6 +22,14 @@ module.exports =
...
           boot()
           juno.runtime.evaluation.eval(move: true)
+      'julia-client:run-cell': =>
+        @withInk ->
+          boot()
+          juno.runtime.evaluation.eval(cell: true)
+      'julia-client:run-cell-and-move': =>
+        @withInk ->
+          boot()
+          juno.runtime.evaluation.eval(cell: true, move: true)
```
pfitzseb: If you implement it like this, the cursor will move to the end of the next *block*, not of the next *cell* like one would probably expect.

lib/runtime/evaluation.coffee
```diff
@@ -16,19 +16,21 @@ module.exports =
     edpath = editor.getPath() || 'untitled-' + editor.getBuffer().id
     fn {editor, mod, edpath}
 
-  eval: ({move}={}) ->
+  eval: ({move, cell}={}) ->
```
pfitzseb: I kinda feel like cell evaluation is sufficiently different than selection/line eval, especially since the `move` behaviour needs to be different as well. Might make sense to create a new method, and maybe factor out the common parts?

lib/package/config.coffee
```diff
@@ -51,6 +51,11 @@ config =
     description: "Limits the Console history's size."
     default: 10000
     order: 8
+  cellDelimiter:
+    type: 'string'
+    default: '#---'
```
pfitzseb: Maybe make the default `'#---|##` or similar? The latter would be pretty much exactly like matlab, which might be a good default.

lib/misc/cells.coffee
```diff
@@ -0,0 +1,32 @@
...
+    if cursor.row > 0
+      buffer.backwardsScanInRange regex, [start, cursor], ({range}) ->
+          start = range.start
+
+    buffer.scanInRange regex, [cursor, end], ({range}) ->
+      end = range.start
+
+    celltext = ed.getTextInBufferRange([start, end])
+    res =
+      range: [values(start), values(end)]
```
pfitzseb: Aren't `start` and `end` just `Point`s and as such already compatible with any API that expects a `[row, column]` array?

lib/runtime/evaluation.coffee
pfitzseb: Or not. Would possibly be cleaner to have a common API for `blocks` and `cells` and then either call the methods from one or the other, i.e. 
```
selector = if cell? then cells else blocks
selector.get(editor)
...
selector.moveNext editor, ...
```

lib/package/commands.coffee
```diff
@@ -22,6 +22,14 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.eval(move: true)
+      'julia-client:run-cell': =>
```
pfitzseb: Should also think about default shortcuts for this.

lib/package/commands.coffee
JobJob: Thanks for the review. Will fix up the rest tomorrow. Just quickly on this point, it does move to the start of the next cell, because, as I was pleased to discover, `blocks.moveNext` goes to the end of the `range` provided, which is the end of the cell in this case.

lib/misc/cells.coffee
JobJob: Yeah good pick up, wasn't paying attention there :)

lib/misc/cells.coffee
JobJob: Yeah I thought it was a bit weird at the time, will re-check it.

dgleich:
Thanks! I gave this a try this morning and I think it works really nicely. I like how the results appear between the blocks!

lib/package/config.coffee
dgleich: I agree.. I think having both is nice.

lib/package/commands.coffee
dgleich: Hydrogen has run-cell mapped to Alt-Command-Enter (OSX) and run-cell-and-move to Alt-Shift-Enter (on OSX), I believe... 

I'd guess there are probably people like me who would map run-cell to Command-Enter (the current run-block)... Matlab's defaults are: Command-Enter (run-cell on OSX) and Command-Shift-Enter (run-cell-and-move) and Control-Enter (run-cell on Windows/Linux) and Control-Shift-Enter (run-cell-and-move on Windows/Linux). 

Not sure how to handle the multiplicity of options at the moment, so being consistent with Hydrogen (and consistent with current Juno!) is probably the best short term solution.

pfitzseb: `Atl-C-Enter` and `Alt-Shift-Enter` sound good to me.

lib/misc/cells.coffee
JobJob: It throws an error [here](https://github.com/JunoLab/atom-ink/blob/master/lib/editor/result.coffee#L92) without `values`. I just matched what `blocks.getRange` was returning.

lib/misc/cells.coffee
JobJob: AFAICT the blocks code seems to iterate over selections, which I've avoided for now since it probably adds a fair bit more complexity to this change. If people want us to add that feature later, then we should look into it. I hope that's not a dealbreaker 😟 but I think this is probably useable enough as it stands without handling (multiple) selections and it's reasonably likely a fair few people probably want this feature merged, asap, pretty bad 🔥🔥🔥

lib/package/config.coffee
JobJob: yep done, with a carat on top.

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
...
+      });
+      return [values(start), values(end)];
+    },
+    get: function(ed) {
+      var range, res;
+      range = this.getRange(ed);
+      res = {
+        range: range,
+        selection: ed.getSelections()[0],
+        line: range[1][1] + 1,
```
pfitzseb: `range[1][0]` is the line, while `[1][1]` is the column. the `+1` is also unnecessary, since that is handled in `eval`.

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
...
+        buffer.backwardsScanInRange(regex, [start, cursor], function(arg) {
+          var range;
+          range = arg.range;
+          return start = range.start;
+        });
+      }
+      buffer.scanInRange(regex, [cursor, end], function(arg) {
+        var range;
+        range = arg.range;
+        return end = range.start;
```
pfitzseb: This will attach the result to the next line with a cell delimiter in it, which is fine, but currently block results are displayed *under* that line, which I find confusing. It'd be trivial to add a patch to ink to allow displaying block results above the marker though. 
If you don't want to look into that codebase I can push a commit for that.

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
...
+    getRange: function(ed) {
+      var buffer, cursor, end, regex, regexString, start;
+      buffer = ed.getBuffer();
+      start = buffer.getFirstPosition();
+      end = buffer.getEndPosition();
+      regexString = atom.config.get('julia-client.cellDelimiter');
+      regex = new RegExp(regexString);
+      cursor = ed.getCursorBufferPosition();
+      if (cursor.row > 0) {
+        buffer.backwardsScanInRange(regex, [start, cursor], function(arg) {
```
pfitzseb: You also shouldn't include the cell delimiter line in the range here -- if you do, then previous results will disappear since the cell ranges overlap.

keymaps/julia-client.cson
```diff
@@ -8,6 +8,10 @@
   'cmd-enter': 'julia-client:run-block'
   'shift-enter': 'julia-client:run-and-move'
   'cmd-shift-enter': 'julia-client:run-file'
+  'alt-cmd-enter': 'julia-client:run-cell'
+  'alt-shift-enter': 'julia-client:run-cell-and-move'
+  'alt-shift-j': 'julia-client:next-cell'
+  'alt-shift-k': 'julia-client:prev-cell'
```
pfitzseb: Hm, those are kinda vim-ish... Maybe `C-j C-n` and `C-j C-p`?  (Same below)

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
+(function() {
+  var scopes, values;
+
+  scopes = require('./scopes');
+
+  values = require('underscore-plus').values;
```
pfitzseb: This is kinda unidiomatic -- should be `import { values } from 'underscore-plus'` at the toplevel. Same above.

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
+(function() {
```
pfitzseb: I don't think it's good style to wrap this in a function.

lib/misc/cells.js
```diff
@@ -0,0 +1,64 @@
...
+  var scopes, values;
+
+  scopes = require('./scopes');
+
+  values = require('underscore-plus').values;
+
+  module.exports = {
```
pfitzseb: It's better to prefix functions you want exported with `export`, i.e.
`export function getRange(ed) {...}`.

lib/misc/cells.js
JobJob: I think if the range returned by `cells.getRange` is from the start of the first line below the top cell delimiter to the end of the line above the end delimiter, everything will work nicely, without needing changes to Ink. This makes sense because then the range is, I think, the same as if you just select the code in the lines between the cell delimiters.

Funnily enough having the result appear below the delimiter was natural to me (coming from Jupyter), but I think it makes sense in atom to have it above, to avoid problems as you said with overlapping ranges and preserving results.

keymaps/julia-client.cson
cdsousa: Hi, I've been triyng this PR and I've setup the following keys which has shown to be very intuitive:
```
'alt-enter': 'julia-client:run-cell'
'alt-shift-enter': 'julia-client:run-cell-and-move'
'alt-down': 'julia-client:next-cell'
'alt-up': 'julia-client:prev-cell'
```

lib/misc/cells.js
pfitzseb: That doesn't really work either, because that way you can't add new lines to the end of an already eval'ed cell without having the output somewhere in the middle (same behaviour you already get at the end of a file right now -- should probably also be changed so that the result is always below the last line in the editor or something. Not critical though, imho).

pfitzseb: Hm, good point. Still don't like it because it's different than the usual cell based coding approaches, but whatever ;)

lib/misc/cells.js
JobJob: Couldn't get `import { values } from 'underscore-plus'` to work for me. Pretty much have no idea when it comes to ES6 modules and require. Need my hand held. We don't need scopes there btw.

lib/misc/cells.js
JobJob: Good we're in agreement that it probably doesn't matter that much for now :) Sorry about the delay, will send through a new version either in a few hours or more likely tomorrow (fingers crossed), bit bogged down with some other work.

lib/misc/cells.js
pfitzseb: Probably need `'use babel'` at the top of the file, like e.g. [here](https://github.com/JunoLab/atom-ink/blob/b3b9d15ae6da254bc2bc04c1449500db98520a63/lib/plots/pane.js).

JobJob: Thanks, but couldn't get that to work, solved it by removing dependency on underscore.

lib/misc/cells.js
JobJob: Also couldn't get that to work... I think it's probably fine as is.

JobJob:
Range is now:
Start of line below top delimiter (and/or start of top row of file)
Start of end delimiter

Results appear below the cell like in Jupyter notebook. Ranges don't overlap so they don't get removed. I think it's good.

lib/misc/cells.js
JobJob: I think I must've been using `use 'babel'` instead of `'use babel'`. Updated to the `export function` style now. `getRange` is not exported in the end.

JobJob:
We good here? Squash the commits?

MikeInnes:
This is a high-quality PR! Thanks @JobJob and @pfitzseb for reviewing. As long as @pfitzseb is happy with the behaviour now I'm happy with the way this is set up (I dislike how long `eval` has become, that needs some refactoring, but it's not your problem).

It's probably going to be a little confusing that this doesn't capture stdout or `display` calls like Jupyter does. But I'm not sure of a good solution for that so we can leave it as it is for a first pass.

Do we need to make the cell delimiter an option? Unless there's a good reason otherwise I'd rather pick a convention and stick with it; the downside is source files being incompatible between different environments / setups.

JobJob:
@pfitzeb had to drag me through it a bit, but I think we got there (or thereabouts) in the end 😄

Yeah, it's a good point re incompatible delimiters, of course you could usually just find-and-replace, but I can forsee version control fights. I guess let's just remove the option, and wait and see if someone complains. 

The default has two options anyway `^(##|#---)` I like them both (first one good to type, second one good to look at) and I guess we can always just add more if people start rioting in the streets.

pfitzseb:
Sorry for the late answer, two last days have been kinda busy.

I'm still not quite convinced that the results should go below the delimiter though -- I'm thinking of `##` starting a new cell, not ending the last one. Maybe @MikeInnes has some sort of opinion?

Having the delimiter as an option might be good for hydrogen compatibilty (they use `# %%` and some others), but we could also just include this one by default. Wouldn't really like supporting all of [these](https://github.com/nteract/hydrogen#hydrogen-run-cell) though.

Also some things I'd like to have in the future:
- visual aid for delimiting cells as in matlab (horizontal line above the `##` or something)
- show the results of all blocks in the cell as if you were eval'ing them one at a time, but aggregated at the bottom.

pfitzseb:
Oh, and apart from all that: I still think we should lint our shiny new ES6 code with e.g. [this](https://atom.io/packages/linter-js-standard). Get's rid of all the ugly `;`s and is in general nicer to look at -- any opinions on whether we should enforce that for new code?

JobJob:
1000% agree re-semicolons, hate them, just didn't want to look too much like a maverick on my first PR to this repo. Will remove them with glee if there's consensus.

JobJob:
Quick poll: @cdsousa @dgleich results for the cell above the `##`, or below?

cdsousa:
Yeah, I also see `##` as being part of the cell they start (like a title in markdown), hence previous cell results should appear above them.

MikeInnes:
> I'm still not quite convinced that the results should go below the delimiter though -- I'm thinking of ## starting a new cell, not ending the last one. Maybe @MikeInnes has some sort of opinion?

I agree with this, the gif above seemed surprising to me.

@JobJob given that we've only just begun our foray into the world of ES6 your opinions are welcome. I'm personally happy to nudge people towards the standard style @pfitzseb mentioned, though I wouldn't try to enforce it on PRs until we have a test setup to do that.

pfitzseb:
Should be easy to set up: https://github.com/feross/standard#what-you-might-do-if-youre-clever

JobJob:
The results don't disappear anymore, but consensus seems to be above the line, so ok. 

I'm going to:

1. remove semi-colons (and see what else this linter-js-standard has to say)
1. remove the option to set the delimiter
1. add # %% as another option
1. squash the commits

Then (option 1) I'm going to leave it to you @pfitzseb to patch Ink to allow results to appear above a specific line. Feel free to ask me to make any changes to this code after you've done that.

Alternately (door number 2) as we discussed earlier I can make the result be below the bottom of the line before the delimiter, with the possible disadvantage that the result won't be above the end of the cell if you type more.

MikeInnes:
I actually think (2) would make more sense – if the line isn't part of the cell it probably shouldn't be part of the range. We can customise how the range works on the ink side so that new text is included within it.

JobJob:
Ok all done.

pfitzseb:
Great, and thanks again for the contribution! 🎉

JobJob:
Np, thanks very much for your help and all contributors for this great package, it's really shaping up.

dgleich:
Great work @JobJob and @pfitzseb

ChrisRackauckas:
Can someone please add a section in JunoDocs.jl on using this?

pfitzseb:
https://github.com/JunoLab/JunoDocs.jl/commit/53e690cef25b39ed76efd58edcfaf46626ef40f3, but the build apparently didn't work.

PR #292, opened 2016-12-11
Debug Toolbar

MikeInnes:
A few improvements to the debug ui, including more memorable icons, as well as putting the debug toolbar in the console (base on https://github.com/JunoLab/atom-ink/pull/100).


MikeInnes:
We really need to push these out, so I'm going to display slightly bad form and merge these PRs early. They shouldn't be too controversial given how much they improve the debugging experience, and obviously any post-hoc review is welcome.

pfitzseb:
Looks good in general, but would be nice if the debugger-related buttons were either inactive or hidden when not debugging.

Issue #311, opened 2017-02-14
Feature Request: Smarter Parsing for .jmd files


kescobo:
Related to [this discussion](https://discourse.julialang.org/t/smart-parsing-jmd-files-weave-jl/1877) on Discourse - It would be nice if the "evaluate all" on `.jmd` files would only evaluate code inside annotated code blocks. This way, when returning to a file to continue an analysis, I could simply open the file, evaluate all, and then pick up where I left off. Currently, it requires highlighting and evaluating each code block separately.

I'd be happy to take a stab at this myself, but would need some guidance as I have not done any developing in Atom.

MikeInnes:
This would be a great feature and a welcome addition. In terms of taking a stab at it your best bet is probably to take a look at doing atom development independently of Juno, at least to begin with. The [atom manual](http://flight-manual.atom.io/) is the best resource for getting started, and you can get the text of the editor and start to process out the cells with commands like:

```js
ed = atom.workspace.getActiveTextEditor()
text = ed.getText()
```
which can be run interactively in the devtools. Once you can do some of that, hooking it up to julia-client for the actual evaluation should be reasonably straightforward.

mpastell:
You can have a look at the https://atom.io/packages/language-weave for regex for parsing the code chunks. It would be great to see this feature, but I also don't have time to work on it right now. 

I would appreciate a mention on Weave related issues, I had missed this thread on discourse as well.

mpastell:
@kescobo I realized that it's actually very simple to get the code from a `jmd` file if you use `language-weave`. Here is a simple function that extracts code from all chunks:

```javascript
    parse_chunks(){
        ed = atom.workspace.getActiveTextEditor();
        text = ed.getText();
        lines = text.split("\n");
        N = ed.getLineCount();
        var code = "";

        for (i=0; i<N; i++)
        {
            scopes = ed.scopeDescriptorForBufferPosition([i,0]).scopes;
            if (scopes.length > 1)
            {
                if (scopes[1] == "source.julia")
                {
                    code += lines[i] + "\n";
                    console.log(lines[i]);
                }
            }
        }
        return(code);
     }
```

kescobo:
@mpastell That's great, thanks! 

If I understand correctly, because `language-julia` already tells the editor that code chunks have julia scope, it might just be a matter I can just check for the scope and evaluate if julia? Seems simple enough :-)

[...]

PR #318, opened 2017-03-08
Add 'julia-client:run-all-chunks' for Weave code chunks

mpastell:
Adds command `julia-client:run-all-chunks` to evaluate all code from Weave `jmd` files as suggested in #311. I will add the keybinding in `language-weave` when this gets merged. 

I'm not sure if this is the best implementation, but it was simple to implement and works similarly to `run-file`.

### DIFF
--- a/lib/package/commands.coffee
+++ b/lib/package/commands.coffee
@@ -28,6 +28,11 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.evalAll()
+      'julia-client:run-all-chunks': (event) =>
+          cancelComplete event
+          @withInk ->
+          boot()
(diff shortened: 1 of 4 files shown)


kescobo:
@mpastell Excellent - simplicity seems best in this case. I'm not really competent to code review, but if it works I say ship it :-)

lib/package/commands.coffee
```diff
@@ -28,6 +28,11 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.evalAll()
+      'julia-client:run-all-chunks': (event) =>
+          cancelComplete event
+          @withInk ->
```
MikeInnes: The following lines need to be indented here, to be part of the `@withInk` callback

PR #337, opened 2017-05-06
notification api

pfitzseb:
implements https://github.com/JunoLab/Juno.jl/issues/52.

Needs https://github.com/JunoLab/Atom.jl/pull/83 and https://github.com/JunoLab/Juno.jl/pull/54 to work.
https://github.com/JunoLab/JunoDocs.jl/pull/19 for docs

### DIFF
--- a/lib/runtime/frontend.coffee
+++ b/lib/runtime/frontend.coffee
@@ -2,7 +2,7 @@
 vm = require 'vm'
 
 {client} = require '../connection'
-{selector} = require '../ui'
+{selector, notifications} = require '../ui'
 
 module.exports =
 
@@ -32,6 +32,9 @@ module.exports =
 
       winActive: (id) =>
         @windows.hasOwnProperty id
+
+      notify: (msg) ->
+        notifications.show msg, true
 
     client.onDetached =>
       for id, win of @windows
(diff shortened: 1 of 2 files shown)


pfitzseb:
@MikeInnes Did we figure anything out regarding synced releases of Atom and Julia packages etc? While this PR doesn't strictly require compatible versions, you'd get `Unrecognised message notify` warnings `julia-client` is to old.

MikeInnes:
Nice.

I think the warning is ok, given that `notify` is explicitly called. Not sure what we could do to force syncing here.

pfitzseb:
Well, in general we could add all Julia packages we need to julia-client (or pull them on installation), much like VSCode does it, but I'm not really a fan of that...

Anyways, I'll merge this for now.

MikeInnes:
Yeah, I could see that causing more problems than it solves. I think the current approach of releasing everything independently is working reasonably well.

KristofferC:
> Well, in general we could add all Julia packages we need to julia-client (or pull them on installation), much like VSCode does it, but I'm not really a fan of that...

Why not? Then you can chose exactly what versions to have of everything and update in sync. Depending on packages installed in the users package directory makes it awkward to develop those packages using Juno (as I experienced when breaking Tokenize :P)

PR #343, opened 2017-05-16
Arguments setting

MikeInnes:
Depends on https://github.com/JunoLab/Atom.jl/pull/85

### DIFF
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -41,7 +41,9 @@ module.exports =
       atom.beep()
 
     @onAttached =>
-      @import('args') atom.config.get 'julia-client.juliaOptions.arguments'
+      args = atom.config.get 'julia-client.juliaOptions.arguments'
+      if args.length > 0
+        @import('args') args
 
   deactivate: ->
     @emitter.dispose()
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -40,6 +40,9 @@ module.exports =
       console.error options.detail
       atom.beep()
 
+    @onAttached =>
+      @import('args') atom.config.get 'julia-client.juliaOptions.arguments'
+
   deactivate: ->
     @emitter.dispose()
     if @isActive() then @detach()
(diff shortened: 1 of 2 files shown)


pfitzseb:
Hm, is this really useful? How common is it to want to have the same `ARGS` for multiple sessions? And if that really is the use case, you could just use [`.junorc.jl`](https://github.com/JunoLab/atom-julia-client/commit/4c9a241803accdef9352aed0d5f0d35f08cb2553), right?

MikeInnes:
@aviks suggested this change; it's not common for interactive workloads but perhaps more so for scripts that get deployed to production. I think the idea is that while it may be technically redundant, it's useful enough and non-intuitive enough for users to be useful.

aviks:
So this came about seeing an user write code that will eventually be run on a server, and controlled from an external process (say a cron or batch run). So the code is written to read arguments from the command line. The issue is that when developing the code in Juno, and using Juno's incremental evaluation features, the user needs to either change the `x=ARGS[1]` line,  or write a check and a fallback. The first is annoying, and the second feels redundant. 

a `.junorc.jl` works, but feels like an extra barrier for people used to graphical environments, exactly the kind of users who are otherwise very happy to use Juno. Having some control of the runtime environment seems like  a standard feature of most fulll-featured IDEs. In eclipse, for example, this is a multi-tabbed set of options.... which may overkill for us, but the point is that this is useful in many situations.

pfitzseb:
Yeah, I get why it would be nice to have a way to specify `ARGS`. But imho putting this into the settings isn't the right way -- `ARGS` should be file- (or at least session-) specific, not something hidden you might forget about.

MikeInnes:
I'm inclined to agree, and I would guess that the Eclipse settings are project-specific rather than global. I imagine that as we hit more enterprise-y use cases we'll see more reasons to support something like that.

At the same time, it's hard to see this causing major problems or confusion, even if it's sub-optimal. I think it'd be ok to have this as a short-term hack and we can start design a more general solution as we need it.

pfitzseb:
Yes, this is totally fine as a stopgap. Project specific settings might be one solution to this problem, a `Execute file with ARGS`-command that opens allows you to specify `ARGS` (with caching) might be another.

But yeah, merging this is fine for now.

MikeInnes:
Note to self, but I should check for empty args so that this is more backwards-compatible.

PR #344, opened 2017-05-19
follow new ink api

pfitzseb:
https://github.com/JunoLab/atom-ink/pull/127

### DIFF
--- a/lib/runtime/evaluation.coffee
+++ b/lib/runtime/evaluation.coffee
@@ -33,17 +33,16 @@ module.exports =
             .catch =>
         else
           r = null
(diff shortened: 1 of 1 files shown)


pfitzseb:
The  `Send to console` button kinda breaks lazy results, will have to look into that. For now it might be better to get the stable changes in and leave that button for another PR.

MikeInnes:
Yup, seems sensible.

pfitzseb:
That code now lives in https://github.com/JunoLab/atom-julia-client/tree/sp/sendtoconsole.

pfitzseb:
This isn't really breaking (we didn't use the `.julia` class we set), so I'm going ahead and merge this.

PR #345, opened 2017-05-20
open console in dock by default

### LINKED ISSUE #107: Lost "console-toggle"
With all the changes in the past few weeks I think we lost the ability to _close_ the console panel.

Before `ctrl-`` would toggle the console, so it would open (close) the console if it was closed (open).

I liked being able to close it, can we get that back?

pfitzseb:
This opens the console in the bottom dock by default, but you can put it anywhere you want. Once opened, `C-j C-o` will toggle the parent dock if there is one or do nothing if not (old behaviour).

Fixes #107.

### DIFF
--- a/lib/runtime/console.coffee
+++ b/lib/runtime/console.coffee
@@ -71,16 +71,8 @@ module.exports =
 
   info: (data) -> @c.info data
 
-  toggle: ->
-    dock = atom.workspace.paneContainerForItem(@c)
-    if dock and dock.toggle
-      if dock.isVisible()
-        dock.hide()
-      else
-        dock.show()
-        @c.open()
-    else
-      @c.open()
+  open: ->
+    @c.open()
 
   reset: -> @c.reset()
 
--- a/lib/runtime/console.coffee
+++ b/lib/runtime/console.coffee
@@ -72,11 +72,15 @@ module.exports =
   info: (data) -> @c.info data
 
   toggle: ->
-    d = atom.workspace.paneContainerForItem(@c)
-    if !d
+    dock = atom.workspace.paneContainerForItem(@c)
+    if dock and dock.toggle
+      if dock.isVisible()
+        dock.hide()
+      else
+        dock.show()
+        @c.open()
+    else
       @c.open()
-    if d.toggle
-      d.toggle()
 
   reset: -> @c.reset()
 
--- a/lib/runtime/console.coffee
+++ b/lib/runtime/console.coffee
@@ -54,8 +54,10 @@ module.exports =
     history.read().then (entries) =>
       @c.history.set entries
     @c.toolbar = @toolbar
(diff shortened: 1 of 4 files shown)


MikeInnes:
Good idea. Might be good to delay the release until the next major, though, in case we need patches to older Atom versions.

pfitzseb:
Mh, I'm pretty sure this can be written in a backwards compatible manner; we can drop the `1.17` requirement then.

pfitzseb:
Okay, this should be fine now, I think.

lib/runtime/console.coffee
```diff
@@ -72,11 +72,15 @@ module.exports =
...
   toggle: ->
-    d = atom.workspace.paneContainerForItem(@c)
-    if !d
+    dock = atom.workspace.paneContainerForItem(@c)
+    if dock and dock.toggle
+      if dock.isVisible()
+        dock.hide()
+      else
+        dock.show()
+        @c.open()
```
MikeInnes: There's a [new API](http://blog.atom.io/2017/05/23/docks-deep-dive.html) for this, so it should be possible to do `atom.workspace.toggle(@c)`. I guess you'll want to avoid that for compatibility but might be worth making a note of

MikeInnes:
I tend to focus the console with `c-o`, can I still do that with this? if not it might make sense to use a different keybinding for that.

pfitzseb:
Uh, if you bind `C-o` to `C-j C-o`, then sure :) Otherwise `C-o` is `Application: Open File` for me.

MikeInnes:
Sorry yeah, that's what I meant

pfitzseb:
Okay, so this actually is a toggle now -- if the console isn't visible we show and focus it, but if it is then it'll be closed. I agree it'd be nice to have a `focus-console` command though -- any suggestions for keybindings?

lib/runtime/console.coffee
pfitzseb: Not really, because we need to call `@c.open()` only when opening the dock. If we don't then the console won't be focused.

MikeInnes:
Yeah it'd be nice to keep that behaviour, especially since it matches the keybinding better. Or we could try having it focus, then close if the console is already focused?

``Ctrl-``` is a nice keybinding for console toggle, but seems to general to add here – we could perhaps add it to uber-juno.

pfitzseb:
Okay, not sure what exactly I was thinking there... This last iteration should be fine: Console is opened (or focused, if it's already open) with `C-J C-O` and can be closed with `C-W`. This will to the right thing if it's in a dock (i.e. Atom will remember where the console was opened and re-open it in that dock). The only thing one might want is a shortcut for closing the dock the console resides in -- this can come up if you have multiple tabs in that dock and therefore can't close it with `C-W`. Not sure how relevant that is though.

PR #346, opened 2017-05-29
Layout fix for 1.17

MikeInnes:
Currently the "use standard layout" option never activates. This fixes that by ignoring the three docks.

Eventually we'll want to use docks for our layout anyway, but this gives us a short term fix.


pfitzseb:
👍 
Using Docks by default would be pretty cool

MikeInnes:
Yeah, docks have the right behaviour in a few places that we're currently hacking around.

PR #349, opened 2017-05-30
In-editor Profile Viewer

pfitzseb:
screenshot_20170530_182102

All the complicated stuff was done by @MikeInnes, I just rebased and made some few improvements.

Depends on https://github.com/JunoLab/atom-ink/pull/133 and https://github.com/JunoLab/Juno.jl/pull/60.

### DIFF
--- a/lib/julia-client.coffee
+++ b/lib/julia-client.coffee
@@ -11,7 +11,7 @@ module.exports = JuliaClient =
   runtime:    require './runtime'
 
   activate: (state) ->
-    etch.setScheduler(atom.views);
+    etch.setScheduler(atom.views)
     @requireInk =>
       commands.activate @
       x.activate() for x in [menu, @connection, @runtime]
--- a/lib/runtime.coffee
+++ b/lib/runtime.coffee
@@ -7,11 +7,13 @@ module.exports =
   plots:       require './runtime/plots'
   frontend:    require './runtime/frontend'
   debugger:    require './runtime/debugger'
+  profiler:    require './runtime/profiler'
 
   activate: ->
     @modules.activate()
     @frontend.activate()
     @debugger.activate()
+    @profiler.activate()
 
   deactivate: ->
     mod.deactivate() for mod in [@modules, @console, @frontend, @debugger]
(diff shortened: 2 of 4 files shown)


mkborregaard:
Awesome!

MikeInnes:
Heads up, I rebased this to remove an unrelated change

MikeInnes:
Originally we just used the generic plot pane for this, which I think makes a lot of sense. Is there a reason to have a profiling-specific pane?

pfitzseb:
I found a profiler specific pane pretty nice since it doesn't destroy any plots then. In principle we could just update our plot pane to handle multiple plots, but that seems like to much trouble for the first iteration of this...

Not sure what doesn't work for you when clicking the profiler boxes -- is a wrong file opened or none at all?

That zooming took me forever to get right, but that was mostly because I tried to use random code from the internet instead of thinking for myself :D

MikeInnes:
It just opens a blank `untitled-2fb38...` file for me.

I partly worry that people will get used to having the new pane and we'll have to stick with it even once we have a better solution. But breaking that might be the lesser evil; it doesn't bother me but I don't do much plotting.

pfitzseb:
Ah, yeah, that's #350. Do you think that's blocking? I dunno how many people work regularly work with untitled buffers...

MikeInnes:
Doesn't have to block this pr, would be nice to get it into the release though

pfitzseb:
Alright, done :)

Re new pane vs PlotPane: I really think a new pane item is better for now, the other option is imho to annyoing if you have a plot open.

pfitzseb:
@MikeInnes Do you think this is good to go?

MikeInnes:
Yeah this looks good to me :+1:

mkborregaard:
*runs `Pkg.checkout("atom-julia-client")` *

MikeInnes:
Awesome, thanks for pushing this along

mkborregaard:
It works! Also great that you can zoom and pan with the mouse, jump into files, etc. This is IMO one of the best inventions after someone came up with toasting day-old bread.
Could there be a possibility of opting out of the top boxes where Atom, CodeTools etc deal process and send the command to julia?

pfitzseb:
Yeah, that should basically be the same scrubbing we already do for stacktraces.

PR #352, opened 2017-06-13
add option for multiple threads

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/61

@ChrisRackauckas: Would you mind trying this real quick on Windows? I'd have to reboot to try if this works there ;)

### DIFF
--- a/lib/connection/process/basic.coffee
+++ b/lib/connection/process/basic.coffee
@@ -100,8 +100,12 @@ module.exports =
               interrupt: => @sendSignalToWrapper 'SIGINT', wrapPort
               socket: @clientSocket proc
 
+  nthreadsset: false
+
   get_: (a...) ->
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Works for me.

marius311:
Great to have this, thanks! One comment, I think it would make more sense if the atom-julia-client setting overrode the JULIA_NUM_THREADS environment variable, otherwise its surprising when you change the Atom setting and nothing happens (if you had the env variable set). Of course, it the atom setting were blank, then it could default to JULIA_NUM_THREADS, or lacking that, default to 1.

pfitzseb:
Good point.
The setting now overrides the global envvar: https://github.com/JunoLab/atom-julia-client/commit/dbbc38694322a8a093ee9b151bd1784f293b81e1

PR #363, opened 2017-06-26
Hyperclick

pfitzseb:
hyperclick

Requires [nuclide/hyperclick](https://atom.io/packages/hyperclick) (but nothing breaks without).

Prompted by https://github.com/JunoLab/atom-julia-client/issues/76#issuecomment-311043474 (cc @kindlychung).

TODO:
- [x] adjust regex so more than one match per line is found

### DIFF
--- a/lib/julia-client.coffee
+++ b/lib/julia-client.coffee
@@ -48,5 +48,7 @@ module.exports = JuliaClient =
 
   provideClient: -> @connection.client
 
+  consumeHyperclick: -> @runtime.consumeHyperclick()
+
   config: require './package/config'
   completions: -> @runtime.completions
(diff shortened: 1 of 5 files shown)


Datseris:
i dont get what it does?

ssfrr:
does this still work? I installed hyperclick and when I hold CTRL my cursor turns into a link-following hand, but I don't see anything underlined and clicking doesn't do anything.

seems like a very cool feature!

ssfrr:
whoops, I hadn't imported the Module whose code I was looking at, so it didn't have any valid definitions to jump to. Now it's jumping as expected (though I don't see the underlines, probably a hyperclick issue not a julia-client issue)

PR #373, opened 2017-07-27
bind ctrl-w to deleting a word

iblis17:
It will be nice to have ctrl-w behave like in REPL.


pfitzseb:
`Ctrl-W` is the shortcut for closing pane items, so if the console isn't in a dock I would expect to be able to close it with that. Also `Ctrl-Backspace` seems to be the more common shortcut to me, and that already works.

iblis17:
ha, okay. (I forgot I have changed the behaviour of `ctrl-w` everywhere. It kept killing my pane when i got typo)

pfitzseb:
Thanks anyways :) I'm not opposed to having similar keybindings in the REPL and Juno console, except for cases like this.

PR #394, opened 2017-11-22
WIP: Console overhaul

pfitzseb:
grafik

### Major issues
- [ ] Async printing doesn't hide REPL prompt. Needs `Base` patch (may work with https://github.com/JuliaLang/julia/pull/24809).
- [ ] Autocompletions and non-standard repl modes don't respect currently set module (cf https://github.com/JuliaEditorSupport/julia-vscode/issues/366). Requires `Base` patch.
- [x] Printing without new line from the editor will be overwritten once REPL prompt is redisplayed. This is a special case of the first item, but we can solve this without any changes to `Base`.
- [x] Non-Julia terminal deserialiazation doesn't work properly.
- [x] Toolbar is missing. I think a VSCode style debugging toolbar is our best bet since the other buttons we had are pretty useles imho.

### Minor issues
- [x] No prompt after starting new julia process when using the cycler.
- [ ] In-editor interrupts once again don't work on Windows. Need to use the powershell wrapper again.
- [x] Console initialization is wonky sometimes (e.g. prints `Main> Main> Main>` or similar).
- [x] Doesn't use `script/boot.jl` right now, so installation/precompilation messages aren't shown.
- [x] Cycler integration.
- [x] Module switching hides the currently entered un-evaled input.
- [x] Copy/paste etc. menu items.
- [ ] Link handler doesn't work for multline links (upstream problem).

### Nice-to-have features
- [x] Click handler for file paths in julia terminal.
- [ ] Select ANSI colours (in terminal and editor) based on currently set editor theme.
- [ ] Auto-generate OhMyREPL theme from currently set theme.
- [ ] Hide console internals in error messages.
- [ ] Keybinding for copying the latest REPL input.
- [x] Send-to-console command.
- [ ] Start-in-external-window command which just spawns an external terminal.
- [ ] Docpane integration in `help>` REPL mode.

### Misc
  - [ ] Fix dependencies:
    - [x] Pre-built binaries. Definitely necessary for Windows, but would also be nice for unixes (because those don't necessarily have `python` and a compiler available afaik). Might be possible to use https://github.com/platformio/pty.js, but I haven't had much luck with that yet. Aside: npm and binary management make me miss BinDeps.jl...
    - [ ] Switch to xterm.js 3.x instead of using the custom VSCode branch.
    - [ ] Use node-pty instead of pty.js.
  - [x] ~~Don't pass Atom keybindings to the terminal. Instead catch them earlier and let Atom handle them.~~ I'm not actually convinced we want this. Copy/paste can just be implemented as regular Atom keybindings, and I don't think we need that much else. Not sure though.
  - [x] Deserializing is broken when multiple terminals are active.
  - [x] Integrate progress metre: 
    - Proper solution would probably be a new REPL mode that can notify the frontend whenever Julia is working (this would also be really cool because we could tap into the help system and show Juno's documentation browser instead of the REPL help).
    - Poor man's solution would be to ping the Julia process and if it doesn't respond we show it as busy.
  - [x] Theme the terminal with whatever style Atom is using.
  - [x] Figure out why the cursor is invisible.
  - [x] Better focus handling.
  - [x] Eval-into-module in terminal
  - [x] ~~Excise all console related stuff in Juno.jl etc.~~ We'll keep the old implementation around for a bit longer, I guess.
  - [x] Debugger integration.
  - [x] Better startup experience.
  - [x] Make default shell customizable.
  - [x] Strip ANSI colour codes from inline errors etc.

### Related PRs:
https://github.com/JunoLab/atom-ink/pull/155
https://github.com/JunoLab/Atom.jl/pull/104


pfitzseb:
Update:
grafik

MikeInnes:
https://github.com/JunoLab/atom-ink/pull/155
https://github.com/JunoLab/Atom.jl/pull/104

Datseris:
Will this complete replace the old console? If yes, why?

MikeInnes:
The plan is to make this the default, and we'll stop making improvements to the old one (possibly remove it completely depending on how people react).

The old console is pretty slow, memory hungry and generally janky. With enough resources we could fix those things, but we don't really have those. And the terminal brings a couple of nice benefits, like seamlessly working with the terminal UIs of a lot of packages. There are downsides, but you can always use the Juno editor to get that experience.

Datseris:
I am all up for replacing the old one, since using the new will have so many advantages.

However there is a big "issue" that needs to be solved. For example myself (and I guess other people) use Juno because I work very interactively. I test stuff in the console and if the work I paste them into my main scripts.

I would really dislike having to delete "julia> " every time I paste. Is there a way you can give this experience, ie that showing "julia> " can be turned off if a user wishes so?

MikeInnes:
That doesn't seem like a disadvantage over the current situation, given that you can currently only copy one result at a time (if you do that in the terminal you can just avoid the prompt).

That said it would be a nice feature to recognise text copied from the terminal and remove the prompts and outputs. Or perhaps even turn the outputs into comments, or something. The repl itself already does this so we can borrow that approach.

Datseris:
> That said it would be a nice feature to recognise text copied from the terminal and remove the prompts and outputs. Or perhaps even turn the outputs into comments, or something. The repl itself already does this so we can borrow that approach.
Yes that would also be great!

> That doesn't seem like a disadvantage over the current situation, given that you can currently only copy one result at a time (if you do that in the terminal you can just avoid the prompt).
A minor comment: I use UP, ctrl+a, ctrl+x, ctrl+tab, ctrl+v and I have everything I used in the console magically appear in my script without ever touching the mouse. Your approach would require me to very precisely select text with the mouse and be very careful to not select any character from the "julia> " :P

Datseris:
I wonder how will this affect the custom printing that Juno uses so far. With trees and everything.

pfitzseb:
Alright, this should be in pretty good shape generally -- there are ~2 big issues remaining:
- I can't get non-julia terminals to be persistent due to some deserializing/serializing issues (this might not actually be that bad, dunno).
- Printing from the editor is a bit messed up: If you eval e.g. `print("foo")` then nothing will be displayed in the terminal since the prompt overwrites the message again. We could print a newline before the prompt everytime something is evaled in the editor, but that'd take up a lot of screen space. I'd *like* to monitor `STDOUT` and `STDERR` and only print the prompt to a new line if something got written to them, but couldn't get that to work.

pfitzseb:
> I wonder how will this affect the custom printing that Juno uses so far. With trees and everything.

It shouldn't affect it much -- you still get custom `render`ing in the editor and normal `show` methods are called in the REPL. `render(::Console, ...)` won't be called, obviously, but that should not be much of a loss generally.

MikeInnes:
Quick thoughts after using this for a bit:

* Booting up and shutting down don't currently give any feedback in the terminal.
* I keep seeing `Main> Main> Main> ...` (sometimes this flashes and disappears, sometimes sticks around).
* Not sure about the terminal module ui. Summarising thoughts elsewhere: we already have a good channel for this information, and it's incompatible with displaying alternate modes like help (which hopefully we will be more compatible with in future).
* I get quite a few errors and warnings in the console
* How hard would it be to get this working with the cycler?
* Would be great if we could get plain terminals to boot the login shell.

Generally it's looking good and seems to be working well!

pfitzseb:
> Booting up and shutting down don't currently give any feedback in the terminal.

Would you think printing a "Juno has closed with ..." message to be sufficiently clear?

> I keep seeing Main> Main> Main> ... (sometimes this flashes and disappears, sometimes sticks around).
Not sure about the terminal module ui. Summarising thoughts elsewhere: we already have a good channel for this information, and it's incompatible with displaying alternate modes like help (which hopefully we will be more compatible with in future).

Yeah, I'll get rid of the whole prompt changing shenanigans.

> I get quite a few errors and warnings in the console

Can you show me a screenshot or something? I don't usually get any, IIRC.

> How hard would it be to get this working with the cycler?

Not that hard, I think.

> Would be great if we could get plain terminals to boot the login shell.

What do you mean by that? They should open to whatever you have selected as `Terminal` in the julia-client settings.

Datseris:
What `checkouts` do I need to do to test this, or just get a taste of it?

pfitzseb:
You need the `sp/terminal` branches for julia-client, ink, and Atom.jl, run `apm install` for the Atom packages, restart Atom, and then select `New Console` as the `Boot Mode` in the Julia client settings.

Datseris:
Oh wait, does this mean that the *old* console that I use e.g. now will **not** be removed? That would be amazing!!!!

pfitzseb:
Yeah, we'll keep that around for now.

pfitzseb:
New debug UI:
grafik

pfitzseb:
Okay, I'm pretty happy with this now apart from some annoying windows stuff (namely interrupts not working) :)
Would be cool if you could test this once again, @MikeInnes. I thought of merging this tonight(ish) so we can hopefully still get a release out this year.

MikeInnes:
So on my system this is booting `sh-3.2 $`. I don't see a terminal option in the settings (does this show up on mac?) but other atom terminals I've used launch zshell correctly. Does that get the issue across? Since you're on windows I guess I may need to do some debugging on my own system to get this working?

By the way, is this the default setup now? Do you think we should release as default straight away or have a round of beta feedback? Once we switch we will need to make sure commands and settings are up to date (e.g. result display mode).

pfitzseb:
Doh... Yeah, that setting isn't available on Mac for some reason. I'll enable it and then you should be able to set it without issues. Could also make `ENV["SHELL"]` the default on unixes if you think that's a good idea?

And yeah, this is the default setup right now, along with Cycler mode. I can revert that for the release, but afaik all settings etc should work fine with the new terminal (except for the scrollback limit I think).

MikeInnes:
> Could also make ENV["SHELL"] the default on unixes if you think that's a good idea?

Yeah, that sounds like the way forward.

Releasing as the default sounds good, but in that case let's have a little beta testing for a week or so to shake out the last issues.

pfitzseb:
> little beta testing for a week

Definitely.

PR #404, opened 2018-01-11
unify copy and interrupt keybindings

### LINKED ISSUE #403: Swap copy/interrupt keybindings on windows/linux
Intercept Ctrl-C & Ctrl-Shift-C via the custom keyhandler on the xterm.js level and switch them around. Then:

| keybinding   | editor    | Juno REPL | standalone REPL |
|--------------|-----------|-----------|-----------------|
| Ctrl-C       | copy      | copy      | interrupt       |
| Ctrl-Shift-C | interrupt | interrupt | copy (windows)  |
| Ctrl-V       | paste     | paste     | paste (windows) |

pfitzseb:
fixes #403.

This has the added benefit that you can use one keybinding everywhere and it will do the smart thing and interrupt either in-editor or in the repl, depending on which is currently running something.

@MikeInnes: I'm not sure if this works for macs, so would be great if you could check.

### DIFF
--- a/lib/connection/process/basic2.js
+++ b/lib/connection/process/basic2.js
@@ -35,6 +35,8 @@ export function get_ (path, args) {
   }
 }
 
+// TODO: reduce code duplication in getUnix and getWindows
+
 function getUnix (path, args, env) {
   return new Promise((resolve, reject) => {
     tcp.listen().then((port, server) => {
@@ -56,6 +58,8 @@ function getUnix (path, args, env) {
             ty: ty,
             kill: () => ty.kill(),
             interrupt: () => ty.kill('SIGINT'),
+            // more robust REPL-only interrupt, in theory only needed on windows
+            // without powershell wrapper:
             interruptREPL: () => ty.write('\x03'),
             socket: sock,
             onExit: (f) => ty.on('exit', f),
@@ -101,6 +105,8 @@ function getWindows (path, args, env) {
                 ty.kill()
               },
               interrupt: () => sendSignalToWrapper('SIGINT', wrapPort),
+              // more robust REPL-only interrupt, in theory only needed on windows
+              // without powershell wrapper:
               interruptREPL: () => ty.write('\x03'),
               socket: sock,
               onExit: (f) => ty.on('exit', f),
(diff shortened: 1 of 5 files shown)


lib/runtime/console2.js
```diff
@@ -53,7 +53,28 @@ export function activate (ink) {
...
+        }
+        return e
+      })
+    } else if (
+      terminal.attachCustomKeyEventHandler((e) => {
+        if (e.ctrlKey && e.keyCode == 67) { // ctrl-c
+          return false
+        }
+        return e
+      })
```
MikeInnes: Did you mean to use `else if` here?

lib/connection/process/basic2.js
```diff
@@ -57,6 +57,7 @@ function getUnix (path, args, env) {
             ty: ty,
             kill: () => ty.kill(),
             interrupt: () => ty.kill('SIGINT'),
+            interruptREPL: () => ty.write('\x03'),
```
MikeInnes: Why do we need this?

lib/runtime/console2.js
pfitzseb: no :)

lib/connection/process/basic2.js
pfitzseb: We don't really except for the case of running on windows without the powershell wrapper.

pfitzseb:
We can't do this with keybindings alone because the xterm.js will process all events before bubbling them upwards for the Atom keyhandlers. ^C (and ^V) is definitely processed by the terminal app, so we need to catch those *before* they reach the terminal.

lib/connection/process/basic2.js
MikeInnes: Ok cool. I think a quick note here would be useful so no-one removes it by accident

MikeInnes:
Oh, one other thing -- how are we handling non-julia terminals? Presumably we should be consistent with them to, it looks like this currently only applies to the julia one

pfitzseb:
Yeah, this is only for the Julia terminal.
And I do think it's better this way -- you basically have one global keybinding for interrupting Julia (Ctrl-Shift-C), but every other terminal isn't affected by that at all.
We could have an option for disabling the special handling of `Ctrl-C` and `Ctrl-V` in the Julia terminal I suppose.

ChrisRackauckas:
Is this also going to catch the fact that console clearing and process killing doesn't work when the console is highlighted?

pfitzseb:
No. Would probably be good to override those though.

pfitzseb:
Alright, this handles ^J and ^K as well, so you can clear the console and kill julia with the normal keybinding while the repl has focus.

PR #467, opened 2018-05-15
Change link to installation instructions in the docs

Karsten7:
The maintained installation instructions seems to be at http://docs.junolab.org/latest/man/installation.html or https://github.com/JunoLab/JunoDocs.jl/blob/master/docs/src/man/installation.md, but not at https://github.com/JunoLab/uber-juno/blob/master/setup.md.


pfitzseb:
Indeed, thanks again!

PR #479, opened 2018-06-22
Optionally allow for static port in "Connect External Process"

tbenst:
Per [this thread](https://discourse.julialang.org/t/connecting-juno-to-work-with-julia-installed-on-a-remote-server/7494/14), it would be nice to have the option to use a static port for use with persistent ssh connections. This pull request implements this feature.


pfitzseb:
Looks great, thanks!

PR #512, opened 2018-09-07
basic HTML pane implementation

pfitzseb:
Requires https://github.com/JunoLab/atom-ink/pull/174.

With those PRs you can call e.g. `
```julia
Atom.msg("newpane", "id", "data:text/html,<div>Hi there</div>", Dict(:icon => "graph", :title => "Foo", :split => "right"))
```
and that will open a new pane (if there's none open with that `id`).

### DIFF
--- a/lib/runtime/plots.coffee
+++ b/lib/runtime/plots.coffee
@@ -3,6 +3,7 @@
 
 {webview} = views.tags
 
+
 module.exports =
   activate: ->
     client.handle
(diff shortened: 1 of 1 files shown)


pfitzseb:
For rendering WebIO widgets:
```julia

function newPane(x, opts=Dict())
    Atom.msg("newpane", "id", string("data:text/html,", sprint(io -> show(io, "application/juno+plotpane", x))), opts)
end

using Interact, Mux
x = filepicker()
newPane(x)
```

PR #528, opened 2018-11-07
toolbar: save icon name updated to sync with fa's update

asbisen:
It seems like font awesome has updated the name of "Floppy Icon" from `floppy-o` to `save`. And that caused the icon to disappear from the toolbar. Here is a fix for that.


pfitzseb:
Thanks for the patch!

PR #531, opened 2018-11-22
Native remote connection support

pfitzseb:
Still lots to do...

- [x] Stop storing passphrase/password in plaintext
- [x] Don't try to `cd` into a local path on startup
- [x] Support for multiple servers.
- [x] Fix interrupts. Kinda works now, but a general solution probably isn't possible since many ssh servers don't implement signal forwarding apparently. 
- [x] Error handling!
- [x] Option for persistent julia process on server.
- [x] Disconnect handling.
- [x] Fix path issues on remote machine
- [ ] Better disconnect handling.
- [ ] Support for more complex ssh setups.
- [ ] Per-server config.

Also needs https://github.com/JunoLab/atom-ink/pull/184 and https://github.com/h3imdall/ftp-remote-edit/pull/228.

### DIFF
--- a/lib/connection.coffee
+++ b/lib/connection.coffee
@@ -24,6 +24,7 @@ module.exports =
 
   consumeInk: (ink) ->
     @IPC.consumeInk ink
+    @ink = ink
 
   consumeTerminal: (term) ->
     @terminal.consumeTerminal term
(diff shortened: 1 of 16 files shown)


pfitzseb:
Alright, this seems to work perfectly on all OSs I have access to. Would be cool to figure out whether this works on MacOS, but I don't think there's a reason it shouldn't.
I'm working on improving the UX a bit (see the checklist above), especially in regards to server management, but feel free to suggest improvements :)

MikeInnes:
I'd like to see an alternative to the global boot setting for this; my best idea is to have a plain command, but we might be able to do something fancier / more automatic.

Also, in normal SSH sessions I rely on key forwarding (e.g. for github access), so that'd be nice to have.

MikeInnes:
Works great!

pfitzseb:
Alright, I'll merge this now and try to fix the last couple of points tomorrow, but this should work well enough regardless.

PR #560, opened 2019-05-28
Recover `focus-last-editor/terminal` functionality

aviatesk:
The commit [add `return-from-goto` command](https://github.com/JunoLab/atom-julia-client/commit/6223ca68efd8fc1832d4321844858808b97eaa3a) removes the functionality for `julia-client:focus-last-editor` and `julia-client:focus-last-terminal`.

I'm not sure this is an intentional change, but I believe this functionality should be so much useful for some users that I recovered it.

I also refactored the file a bit with following changes:
- Add identifiers to variables (suggested by eslint)
- Removed unused functions (grep-based check for usages in the whole project)
- Use Atom `CompositeDisposable` for disposer

### DIFF
--- a/lib/ui/focusutils.js
+++ b/lib/ui/focusutils.js
@@ -4,9 +4,8 @@ import {TextEditor} from 'atom'
 
 let lastEditor
 let lastTerminal
-let listener
+let listner
 let commands
-let ink
 
 function clamp (x, min, max) {
   return Math.min(Math.max(x, min), max)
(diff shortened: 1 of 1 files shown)


lib/ui/focusutils.js
```diff
@@ -50,48 +44,42 @@ class FocusHistory {
...
-  let history = new FocusHistory(30)
+  subs.add(atom.workspace.onDidStopChangingActivePaneItem((item) => {
+    if (item instanceof TextEditor) {
+      lastEditor = item
+    } else if (item instanceof i.InkTerminal) {
+      lastTerminal = item
+    }
+  }))
 
+  const ink = i
```
pfitzseb: Don't need this anymore -- should just change the `i` function argument to `ink` and that's it.

aviatesk: Ah I missed to refactor it. Fixed !

PR #561, opened 2019-06-04
Feature - Add custom layout options

aviatesk:
### Rationale

Even though Juno's serialization works well, we still have to manually layout various panes at the beginning of the project or the later.
This PR addresses this by adding custom layout options so that users can set up their own default layout, i.e.: the default location and the splitting direction of any Juno pane.

### Approach

The possible drawback will be that these settings can be messed up. Thus I took the two approaches below:
1. Set the default settings to such values that they behave _exactly_ the same way as before
2. Made these settings collapsed by default so that users won't feel bother at them. Only users who want to tweak the default pane-opening settings, they can as they want.

Furthermore, I use [this new enum config settings](https://github.com/atom/settings-view/pull/1089) and I think these config seems rather neat:
image

### Further Idea

If this custom layout options are approved, I also want to implement an optional functionality like `Restore Standard Layout on Start`:
It would be like "Opens user-selected panes on their desired location on startup except deserialized ones" and I believe this is much useful for some users.

### Notes

- [This](https://github.com/JunoLab/atom-ink/pull/203) is the PR on atom-ink side: It doesn't include any kind of atom-julia-client specific changes
- I additionally changed the variable-identifiers in files I modified to appropriate ones (like `let` to `const`). But if you feel bother about this, I can easily revert the changes since I put together those changes into the last commit.

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -150,12 +150,178 @@ config =
         type: 'boolean'
         default: false
         order: 9
+      customLayoutOptions:
+        title: 'Custom Layout Options'
+        type: 'object'
+        order: 10
+        collapsed: true
+        properties:
+          console:
+            title: 'Console'
+            type: 'object'
+            order: 1
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Console Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'bottom'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Console Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
+          terminal:
+            title: 'Terminal'
+            type: 'object'
+            order: 2
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Terminal Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'bottom'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Terminal Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
+          workspace:
+            title: 'Workspace'
+            type: 'object'
+            order: 3
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Workspace Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'center'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Workspace Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'right'
+                radio: true
+                order: 2
+          documentation:
+            title: 'Documentation Browser'
+            type: 'object'
+            order: 4
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Documentation Browser Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'right'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Documentation Browser Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
+          plotPane:
+            title: 'Plot Pane'
+            type: 'object'
+            order: 5
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Plot Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'center'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Plot Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'right'
+                radio: true
+                order: 2
+          debuggerPane:
+            title: 'Debugger Pane'
+            type: 'object'
+            order: 6
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Debugger Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'right'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Debugger Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
+          profiler:
+            title: 'Profiler'
+            type: 'object'
+            order: 7
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Profiler Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'center'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Profiler Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'right'
+                radio: true
+                order: 2
+          linter:
+            title: 'Linter'
+            type: 'object'
+            order: 8
+            collapsed: true
+            properties:
+              defaultLocation:
+                title: 'Default location of Linter Pane'
+                type: 'string'
+                enum: ['center', 'left', 'bottom', 'right']
+                default: 'bottom'
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Linter Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
       noAutoParenthesis:
         title: 'Don\'t Insert Parenthesis on Function Autocompletion'
         description: 'Juno will not insert parenthesis after completing a function if this is enabled.'
         type: 'boolean'
         default: false
-        order: 10
+        order: 11
   consoleOptions:
     type: 'object'
     order: 4
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -226,14 +226,14 @@ config =
                 title: 'Default location of Documentation Browser Pane'
                 type: 'string'
                 enum: ['center', 'left', 'bottom', 'right']
-                default: 'right'
+                default: 'center'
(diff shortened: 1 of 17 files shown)


lib/package/config.coffee
```diff
@@ -150,12 +150,178 @@ config =
...
+                radio: true
+                order: 1
+              split:
+                title: 'How to split Linter Pane'
+                type: 'string'
+                enum: ['no split', 'left', 'up', 'right', 'down']
+                default: 'no split'
+                radio: true
+                order: 2
       noAutoParenthesis:
```
aviatesk: I think this setting is kind of `Julia Options` rather than `UI Options`.
Stays here for consistency for the past version ?
@pfitzseb

pfitzseb: Yeah, probably makes sense to move it. Should probably update [`updateSettings`](https://github.com/JunoLab/atom-julia-client/blob/master/lib/package/settings.js) to be recursive and clean out the old config (there's a problem where old settings are still shown if the user has changed it).

aviatesk: Ah I've not noticed the file. Will take a look at and modify it.

aviatesk: @pfitzseb I've implemented recursive version of `updateSettings` 😉

aviatesk:
## Achievements

- [x] The default location and splitting rule of each Juno pane is now configurable
- [x] Update `Julia:Standard-Layout`. It's now renamed to `Julia-Client:Default-Layout` and this nicely reflects the user-defined settings. See [lib/ui/layout.js](lib/ui/layout.js)
    * [x] Update `queryStandardLayout` according to the change above
- [x] Update configs settings
- [x] Implement recursive version of [`updateSettings`](lib/package/settings.js#L15). Now it detects all the invalid (deprecated) config settings and notify them to an user and ask her whether or not to remove them

## Quick look

Config settings:
image

Atom workspace after invoking `Julia-Client:Default-Layout` with default config values:
image
(This is not a big deal, but actually I personally think the default location of `Workspace` and `Plot Pane` is better to be `"right"` dock as `Documentation Browser` is. If you think so as well, I want to change the default value of them)


## Notes

- This commit includes various internal refactors, like changing variable identifiers, modify exported functions/variables, rename the function, etc. But I carefully checked every feature stays working after these refactors.
- This commit updates various configs, and not totally compatible to the previous version. (Thus should bump the minor version ?)

lib/ui/layout.js
```diff
@@ -1,101 +1,87 @@
...
-  for (let pane of [workspace(), cons(), docs(), plots()]) ps.push(pane.close())
-  Promise.all(ps).then(() => {
-    resetLayout()
-    createLayout(layout)
-    atom.commands.dispatch(atom.views.getView(atom.workspace.getActivePane()), 'julia-client:open-console')
-    if (centralpane.activate) centralpane.activate()
+    ],
+    description:
+      `You can specify the panes to be opened and their _default location_ and _splitting rule_ in
+       **'Settings -> Julia-Client > UI Options > Layout'**.\n
```
pfitzseb: Maybe  put the settings path in backticks?

lib/ui/layout.js
```diff
@@ -1,101 +1,87 @@
...
 }
 
-function workspace() { return require('../runtime').workspace.ws }
-function docs() { return require('../ui').docpane.pane }
-function cons() {
-  if (atom.config.get('julia-client.consoleOptions.consoleStyle') == 'REPL-based') {
-    return require('../runtime').console2.terminal
-  } else {
-    return require('../runtime').console.c
+function openPanes (panes) {
```
pfitzseb: What happens if an unexpected pane item is focussed when this runs?

lib/package/settings.js
```diff
@@ -1,18 +1,80 @@
...
 }
+
+/**
+ * Recursively search deprecated configs
+ */
+function searchForDeprecated(config, currentSchemes) {
+  Object.entries(config).forEach(([key, value]) => {
+    // @NOTE: Traverse the current config schemes by post-order in order to push all the invalid
+    // config schemes into `invalidSchemes`
+    if (Object.prototype.toString.call(value) === '[object Object]') {
```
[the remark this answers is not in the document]
aviatesk: `value` can be an array and in that case `typeof value === "object"` won't work, cause `typeof ary === "object"`

aviatesk:
@pfitzseb Thanks for your review ! And I hope you would like this feature 😄

lib/ui/layout.js
aviatesk: I tried that with `[**'Settings -> Julia-Client > UI Options > Layout'**](atom://config/packages/julia-client)]` but that didn't work 😢 
Any other idea ?

pfitzseb: You could add a button that opens the settings, but I don't think that's worth it. Imho 
```
`Settings -> Julia-Client > UI Options > Layout`.\n
```
would just look better, but it's not super important.

lib/package/settings.js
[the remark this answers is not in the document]
pfitzseb: Ah yes, the joys of type checking in JS. Nevermind me ;)

lib/ui/layout.js
aviatesk: Umm could you tell me an idea about an example of "an unexpected paen item" ??

lib/ui/layout.js
aviatesk: Yeah, I also wondered about adding the button, but I thought that would make the notification looks more complicated, thus I didn't do that.

If you're okay, I want to let this go for now.

aviatesk:
@pfitzseb 
btw, how about this ??

> (I'm not sure about what default settings are the best, but personally think the default location setting of Workspace and Plot Pane is better to be "right" dock as Documentation Browser is. If you think so as well, I want to change the default value of them)

Your `standardLayout` lays Documentation Browser in a center pane, and so I can follow that if you want instead of my idea.

lib/ui/layout.js
pfitzseb: Peek 2019-06-07 20-55

In that gif the focus initially is on the editor in the bottom right, which leads to a kinda funky layout. I think the previous implementation closed all panes first and then started opening the Juno panes.

pfitzseb:
Yea not sure about what the standard layout should be. I think for now we should probably stick with the current one.

Also, do you by chance have access to an OSX system? IIRC there were a lot of problems with that before.

lib/ui/layout.js
pfitzseb: Yeah, let's keep it like this for now.

aviatesk:
Okay, I will change the `defaultLocation` of Documentation Browser to `"center"` !
And unfortunately I'm just using Windows 10 for now. You mean "problems" are these layout stuff ??

lib/ui/layout.js
aviatesk: Daaaaamn, seriously funny state...
I thought not closing all the pane would be better in a sense that even with this command an user can keep a part of the previous state, but sure, closing all the panes seem better....

pfitzseb:
Ok, so this is almost good to go. The one problem right now is that `defaultLayout()` doesn't reset the current layout. So if I have the default location of the REPL set to bottom but it's actually open in the workspace center, running `Julia Client: Default Layout` doesn't put it in the right place (which is different from the old `Standard Layout` command).
Would be awesome if you could take a look at fixing that.

aviatesk:
Understood, I'm going to implement the fix until tomorrow 👍

pfitzseb:
No hurry :) Thanks for the PR, again! :)

aviatesk:
@pfitzseb 
I've just implement the fix the above problem ! Please check it and get me back your feedback :)

One thing I'm concerning is I occasionally get the error below in dev console and then this command appears to only close the Juno panes:
```
Uncaught (in promise) Error: The workspace can only contain one instance of item [object Object]
    at ItemRegistry.addItem (C:\Users\aviat\AppData\Local\atom\app-1.37.0\resources\app\static\<embedded>:14:41773)
    at PaneContainer.didAddPaneItem (C:\Users\aviat\AppData\Local\atom\app-1.37.0\resources\app\static\<embedded>:11:446465)
    at Pane.addItem (C:\Users\aviat\AppData\Local\atom\app-1.37.0\resources\app\static\<embedded>:11:458916)
    at Pane.activateItem (C:\Users\aviat\AppData\Local\atom\app-1.37.0\resources\app\static\<embedded>:11:457779)
    at Workspace.open (C:\Users\aviat\AppData\Local\atom\app-1.37.0\resources\app\static\<embedded>:11:431512)
    at <anonymous>
```

If I execute the command again it works, but should I have to approach this issue in this PR ?
I think this is not specific to this command, but opening/closing `PaneItem` instances in general.

lib/ui/layout.js
aviatesk: Ah, I totally misunderstood what you meant ! Backticks !
Yeah, it should look better, thus I changed it in the recent commit 😉

lib/ui/layout.js
```diff
@@ -1,101 +1,139 @@
...
+  if (atom.config.get('julia-client.uiOptions.layouts.defaultPanes.linter')) panes.push(linter)
+
+  return panes
 }
 
-export function isCustomLayout() {
-  return getPanes().length > 1
+function closePromises () {
+  // Close only specified panes, i.e.: non-specified panes won't layouted
+  const panes = specifiedPanes()
```
pfitzseb: I think it would be best to flatten all items in the workspace center into one pane and then only closing Juno pane items that are in the workspace center. After that you can focus the one remaining pane in the workspace center and start splitting stuff from there.

But the current implementation is fine too, so no need to change that if you don't want to.

aviatesk: In the current implementation, I does that too.
The function `bundlePanes` flattens all the items in each container, cause now we have an option to open Juno panes outside of `workspaceCenter` (and without this, say, right dock can be funky state after the command as well like you showed yesterday)

pfitzseb: Right, but it's much less likely to have splits in docks (at least for me).
We could also specialcase the no-splits case and don't close/reopen stuff that's in the right place already.

aviatesk: > We could also specialcase the no-splits case and don't close/reopen stuff that's in the right place already.

IMO we better to close/reopen even with `no-split`, cause a pane with `split === 'no split'` can be outside of the `defaultLocation` and it should be moved into the right location.

pfitzseb: Sure, we'd need to check that each pane item is in the correct place (i.e. workspace center or dock), that there are no splits in that pane, and that the user doesn't want any splits for those items. But that's probably not worth the effort.

aviatesk: Ah I see. Agree with you that I also wonder it's worth doing. So let this go for now.

PR #562, opened 2019-06-07
Add a command to toggle conditional breakpoint

aviatesk:
This PR enables an user to easily create a conditional breakpoint.

Behaviour of `toggle-conditional-breakpoint`:
- If there is no breakpoint: Create a brand new conditional breakpoint
- If there is already a non-conditional/conditional breakpoint: Edit its condition

The corresponding PR in atom-ink: https://github.com/JunoLab/atom-ink/pull/204

### DIFF
--- a/lib/package/commands.coffee
+++ b/lib/package/commands.coffee
@@ -41,9 +41,9 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.eval(cell: true, move: true)
-      'julia-client:next-cell': ->
+      'julia-client:next-cell': =>
         cells.moveNext()
-      'julia-client:prev-cell': ->
+      'julia-client:prev-cell': =>
         cells.movePrev()
       'julia-client:goto-symbol': =>
         @withInk ->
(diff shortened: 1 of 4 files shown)


lib/package/menu.coffee
```diff
@@ -51,9 +51,10 @@ module.exports =
 
       {type: 'separator'}
 
-      {label: 'Open Plot Pane', command: 'julia-client:open-plot-pane'}
       {label: 'Open Workspace', command: 'julia-client:open-workspace'}
       {label: 'Open Documentation Browser', command: 'julia-client:open-documentation-browser'}
+      {label: 'Open Plot Pane', command: 'julia-client:open-plot-pane'}
```
pfitzseb: Any reason you changed the ordering here?

lib/package/commands.coffee
```diff
@@ -41,9 +41,9 @@ module.exports =
         @withInk ->
           boot()
           juno.runtime.evaluation.eval(cell: true, move: true)
-      'julia-client:next-cell': =>
+      'julia-client:next-cell': ->
```
pfitzseb: I'd prefer keeping these as fat arrows (just for consistency with the JS code).

aviatesk: Sure I revert the commit soon 😉

lib/package/menu.coffee
aviatesk: This is because in my custom-layout PR, the order of those Juno panes are:
Console -> Terminal -> Workspace -> Documentation Browser -> Plot Pane -> Debugger Pane -> Profiler -> Linter

When deciding the order, I naively refered to these orders: https://github.com/aviatesk/atom-julia-client/blame/master/lib/runtime.coffee#L5

pfitzseb: Ok, fair enough.

PR #565, opened 2019-06-16
Rename layout command

aviatesk:
I want to rename these layout-related command so that they are named as _verb-prefixed_, because most commands in Julia-Client are prefixed by verbs, like `Julia-Client:Open-Plot-Pane`.

This PR also includes a minor change in `restoreDefaultLayout`.


pfitzseb:
Looks good, thanks!

PR #566, opened 2019-07-08
Create command to format code in a current editor

aviatesk:
## Purpose

This PR adds `Julia-Client: Format-Code` command to format code in a currently opened editor.

The corresponding PR on Atom.jl side is: https://github.com/JunoLab/Atom.jl/pull/142

## Misc

This PR also clean up command/keymap registries. This is basically done in [this commit](https://github.com/JunoLab/atom-julia-client/commit/b1c5cf24db371d5521a9156b59941662f2641e05) so please refer to the commit message and the changes

### DIFF
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -175,7 +175,7 @@ module.exports =
   connectedError: (action = 'do that') ->
     if @isActive()
       atom.notifications.addError "Can't #{action} with a Julia client running.",
-        description: "Stop the current client with Packages → Julia → Stop Julia."
+        description: "Stop the current client with `Packages -> Julia -> Stop Julia`."
       true
     else
       false
@@ -183,7 +183,7 @@ module.exports =
   notConnectedError: (action = 'do that') ->
     if not @isActive()
       atom.notifications.addError "Can't #{action} without a Julia client running.",
-        description: "Start Julia using **`Julia-Client: Start-Julia`** command."
+        description: "Start a client with `Packages -> Julia -> Start Julia`."
       true
     else
       false
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -183,7 +183,7 @@ module.exports =
   notConnectedError: (action = 'do that') ->
     if not @isActive()
       atom.notifications.addError "Can't #{action} without a Julia client running.",
-        description: "Start Julia using Packages → Julia → Start Julia."
+        description: "Start Julia using **`Julia-Client: Start-Julia`** command."
       true
     else
       false
--- a/lib/runtime/formatter.js
+++ b/lib/runtime/formatter.js
@@ -13,29 +13,47 @@ export function formatCode () {
     formatEditor(editor)
   } else {
     selections.forEach((selection) => {
-      formatSelection(editor, selection)
+      formatEditorWithSelection(editor, selection)
     })
   }
 }
 
 function formatEditor (editor) {
+  const range = editor.getBuffer().getRange()
+  formatEditorTextInRange(editor, range, editor.getText())
+}
+
+function formatEditorWithSelection (editor, selection) {
+  const range = selection.getBufferRange()
+  formatEditorTextInRange (editor, range, selection.getText())
+}
+
+function formatEditorTextInRange (editor, range, text) {
+  const marker = markRange(editor, range)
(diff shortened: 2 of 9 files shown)


lib/connection/client.coffee
```diff
@@ -183,7 +183,7 @@ module.exports =
   notConnectedError: (action = 'do that') ->
     if not @isActive()
       atom.notifications.addError "Can't #{action} without a Julia client running.",
-        description: "Start Julia using Packages → Julia → Start Julia."
+        description: "Start Julia using **`Julia-Client: Start-Julia`** command."
```
pfitzseb: Should probably have both options here or keep this as is, since advanced users will know how to start Julia and everyone else is more likely to use the menus.

lib/runtime/formatter.js
```diff
@@ -0,0 +1,43 @@
+/** @babel */
+
+import { client } from '../connection'
+
+const { format } = client.import(['format'])
+
+
+// @TODO: Once DocumentFormat.jl comes to be able to handle options to specify indents,
+//        we can extract some information from the status-bar tile consumed by atom-indent-detective,
+//        or more safely, make atom-indent-detective provide the service and use it here ?
```
pfitzseb: We can just query `atom.workspace.getActiveTextEditor().getTabLength()`, no?

menus/julia-client.cson
```diff
@@ -9,6 +9,7 @@
         {label: 'Select Block', command: 'julia-client:select-block'}
         {label: 'Go to Definition', command: 'julia-client:goto-symbol'}
         {label: 'Show Documentation', command: 'julia-client:show-documentation'}
+        {label: 'Format Code', command: 'julia-client:format-code'}
```
pfitzseb: Maybe also add an entry to `lib/package/menu.coffee`?

lib/runtime/formatter.js
```diff
@@ -0,0 +1,43 @@
...
+  })
+}
+
+function formatSelection(editor, selection) {
+  const range = selection.getBufferRange()
+  format({
+    text: selection.getText(),
+    // indent: INDENT // @TODO
+  }).then(({ formattedtext }) => {
+    editor.setTextInBufferRange(range, formattedtext)
```
pfitzseb: What happens if I add a new line somewhere else before the `Promise` is fulfilled?
It might make sense to use the marker API here (and we can also add a decoration to show what we're formatting). That'd make it possible to 
- keep track of the selected text no matter what other edits were made to the rest of the document
- don't apply the formatting if the content of the marked range changes

menus/julia-client.cson
pfitzseb: Or even a toolbar button. This feature is something lots of people have asked for, so I feel it's warranted to give it some exposure.

lib/runtime/formatter.js
aviatesk: I thought it would be great if we can use `getIndent` in `atom-indent-detective` on each call of the formatting command, but I think `getTabLength()` API would be simple better.

lib/connection/client.coffee
aviatesk: Sure, I will follow the menus way :smile:

lib/runtime/formatter.js
aviatesk: The current implementation just _overwrites_ the text inside of the range even when it's modified before the `Promise` gets fullfilled.

And I think the marker would be more friendly and nicer, thus I will give a try to it.

Maybe is there some utility for marking a editor like the one we use on code execution ?

aviatesk:
I created a too-bar button for `format-code` and also tweaked some others:

| Before | After |
|---|---|
| image | image |

What do you think on this change ?

lib/runtime/formatter.js
aviatesk: I wondered `ink.highlight` can be useful here, but it always destroy markers after 20ms now, thus I will implement creating newer function in `formatter.js`

pfitzseb: The decoration stuff isn't as important as tracking the right selection and canceling the formatting if the user makes changes, so don't worry about it too much.

aviatesk: I see. I've done the work in the last commit 😄

PR #567, opened 2019-07-08
Fixed line numbers for weave cell evaluation.

twadleigh:
Addresses https://github.com/JunoLab/Juno.jl/issues/325


pfitzseb:
Looks good, thanks!

PR #573, opened 2019-07-19
Add hover

aviatesk:
Refer to [the PR on Atom.jl](https://github.com/JunoLab/Atom.jl/pull/145) for the details.

## TODOs on atom-julia-client side

- [x] Make [the regex to detect breaks of words](https://github.com/JunoLab/atom-julia-client/blob/71c6be2b4fb7fa9d56f6e3d820c910fb48021483/lib/runtime/datatip.js#L15) work with non-ascii characters
    ~~* The kinda negative version of [regex in misc/words.coffee](https://github.com/JunoLab/atom-julia-client/blob/71c6be2b4fb7fa9d56f6e3d820c910fb48021483/lib/misc/words.coffee#L2) didn't work (like `/[^\u00A0-\uFFFF_!´\.@]/`). I'm not super familiar with this kind of stuff and appreciate the help.~~
- [x] Detect modules correctly for non-active editors (like the split one on the right and left)
- [ ] Make config to disable/enable feature and add a notification for how to install Atom-IDE related packages for an user when s/he enable the config setting but not has installed the service package ?

## Notes

~~- The integration is disabled for `comments` and `strings` scopes.~~
- The integration is disabled for `comments` scope and enabled for `string` scope because of the consistency with Auto-completions. And note that they're not active for texts that're not Julia embedded code in .jmd files
    * Hyperclick integration follows the same rule
    * But `Julia-Client: Goto-Symbol` is still dispatchable in any scope of `juliaSyntaxScopes` since it can be invoked by menus

### `goto` feature & its Hyperclick integration

- Separates code from [evaluation.coffee](lib/runtime/evaluation.coffee)
- We consume the latest version of hyperclick-service: `0.1.0`
- Hyperclick now detects modules for non-active editors
- Hyperclick doesn't show underlines when there is no destination to go
- [x] One `@XXX` remains (might be very hyperclick-specific issue and I think we can accept this since it works and there would be sadly no further change to the hyperclick package)
https://github.com/JunoLab/atom-julia-client/blob/71c6be2b4fb7fa9d56f6e3d820c910fb48021483/lib/runtime/goto.js#L84-L91

### JSfies

I JSfied the files below since I feel more controls for JS over coffees and it enables me to develop easily
- words.coffee
- scopes.coffee

### Misc

- Better dispositions of `runtime` and `ui`
- UI-related refactors of `modules.coffee`'s `status-bar` view
    * Update just now change `@dom.style` rather than destroying the item
    * Return back `Disposable` to the `status-bar`


lib/runtime/datatip.js
```diff
@@ -0,0 +1,93 @@
...
+      start -= identifierStart
+    }
+    let end = column
+    const identifierEnd = text.slice(column).search(wordRegex)
+    if (identifierEnd === -1) {
+      end = text.length
+    } else {
+      end += identifierEnd
+    }
+    const word = text.slice(start, end)
```
pfitzseb: It would probably be easiest to do something like [this](https://github.com/atom/atom/blob/v1.38.2/src/cursor.js#L606-L616) with the regex already defined in words.js.

pfitzseb:
Do you know *why* 
```
 One @FIXME remains (might be very hyperclick-specific issue and I think we can accept this since it works and there would be sadly no further change to the hyperclick package)
```
happens? Is it because the editor is focused at some point which closes the overlay or something like that?

aviatesk:
I tried to track the source of problem, but still not sure why this happens.

But here is what I've been able to track:
When we've clicked the mouse, [this part of code in hyperclick package](https://github.com/facebookarchive/atom-ide-ui/blob/master/modules/atom-ide-ui/pkg/hyperclick/lib/HyperclickForTextEditor.js#L200^L210) gets fired and it fails to keep the overlay open without the funky `setTimeout`, but if we use the explicit command `hyperclick: confirm-cursor` rather than clicking, [this part](https://github.com/facebookarchive/atom-ide-ui/blob/master/modules/atom-ide-ui/pkg/hyperclick/lib/HyperclickForTextEditor.js#L364-L370) is fired and it _works_.

The main obvious difference is that the later `await`s for an newer suggestion but the former uses a cache rather than creating an new.
And if I change `_onMouseDown` into `async` function so that it `await`s for an newer suggestion as `confirmSuggestionAtCursor` does, it turns into to work.
But still I'm not sure why this `await` stuff helps here, since the other parts in `_onMouseDown` don't seem to do the focusing stuff as you guess...

lib/runtime/datatip.js
aviatesk: With [this commit](https://github.com/JunoLab/atom-julia-client/pull/573/commits/94641a17415b741f382f87c968e8584295c8e3cc), the regex and index stuff turn to work with non-ascii characters.
Screenshot 2019-07-22 16 18 09

But yes, it would be better to borrow the regex from word.js.
I will change the implementation if I have time.

aviatesk: Yeah, solved in [this commit](https://github.com/JunoLab/atom-julia-client/pull/573/commits/2f938bf31b6f91105cb8fdd0d3d5386aa8a5256d).
Thanks for the suggestion.

aviatesk: image

aviatesk:
Will resolve conflicts and give a another look on [this](https://github.com/JunoLab/atom-julia-client/pull/573#issuecomment-513590540) later.

pfitzseb:
Great. Imho we should recommend `atom-ide-datatip` (and maybe add whatever functionality is missing). Might also make sense to automatically install it in uber-juno.

aviatesk:
> Imho we should recommend atom-ide-datatip (and maybe add whatever functionality is missing).

Agree. I think it would lead to bad things if we rely on UIs of atom-ide-ui. 
I won't add any functionality to Juno that uses atom-ide-ui's services other than this datatip.
(May try to refactor this by creating "hover UI" only for Juno/ink and use that in the future)

> Also, don't worry too much about the setTimeout thing -- it's good enough if it works.

Thanks for kind responses. But let me investigate a bit more.

aviatesk:
@pfitzseb 
Maybe because of Atom's update of its electron, the callback just works now without the weird `setTimeout`.
(I changed the whole file to use ES6 class syntax, but it's not related)

I've not dug into the details of changes between v2 and v3, but I think it would be okay since atom-julia-client is now only compatible with `>=v1.39.0`

pfitzseb:
Alright, I'll go ahead and merge this then. Thanks for the great work!

PR #579, opened 2019-07-27
Pending open

aviatesk:
Introduces `Open files in pending state` config.

Please refer to the description of the config for why I want to introduce this: 
https://github.com/JunoLab/atom-julia-client/pull/579/files#diff-5655a196ecf76f9731b64509e8b7676aR141-147

The corresponding PR on atom-ink is: https://github.com/JunoLab/atom-ink/pull/208

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -140,16 +140,6 @@ config =
         description: 'Show plots in Atom.'
         default: true
         order: 7
-      pendingOpen:
-        title: 'Open files in pending state'
-        description: 'If enabled, Juno will open files
-                      [in a pending state](https://flight-manual.atom.io/api/v1.39.1/Workspace/#instance-open)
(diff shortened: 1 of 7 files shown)


aviatesk:
> I think we should just reuse atom.config.get('core.allowPendingPaneItems') instead of introducing our own config option.

Sounds good. Is it okay that it would change the file-opening behaviour implicitly from the an user's veiw point ?

> IMHO we have too many options as is.

Totally agree. This is just a talk of appearance, but how about collapsing all the top level config like this ?
image

This would make config appear more structured and may navigate an user better.

pfitzseb:
> Sounds good. But is it okay that it would change the file-opening behaviour implicitly from the an user's veiw point ?

Should be ok, imho, especially since the core setting doesn't precisesly specify when it is used.

> This is just a talk of appearance, but how about collapsing all the top level config like this ?

Sure, sounds good.

aviatesk:
Addressed in the commits above 👍

aviatesk:
@pfitzseb Please merge this first before #573

PR #580, opened 2019-07-28
Refactor AutoCompletions:

aviatesk:
- JSfy completions.coffee
- Only loads completions.js when an user consumes Auto-CompletePlus pacakge
- Add config for `suggestionPriority`
- Make `iconHTML` more consistent with Documentation/Workspace


lib/package/config.coffee
```diff
@@ -77,12 +77,23 @@ config =
...
+        title: 'Auto-Completion Suggestion Priority'
+        description:
+          '''
+          Specify the sort order of Auto-completion suggestion from Julia-Client.
+          Note the default providers like snippets have priority of 1.
+          Requires Atom restart to take an effect.
+          '''
+        type: 'integer'
+        default: 3
+        order: 11
```
pfitzseb: Does this really need to be a setting?

aviatesk: I personally uses snippets a lot and IMHO I want to have a control over priorities of completions like this.

But at the same time I doubt that most users have only 2 completion providers for Julia files (from Juno and snippet), and thus I this we can make this config into `boolean` one, which would appear simpler to user's view point.

pfitzseb: Fair. Let's keep it like this then.
I personally have snippets disabled all the time, but I guess giving the julia-client completions a higher priority would be even better.

aviatesk: Cool. Maybe the default value is better to be more than `3`, but who knows.

PR #585, opened 2019-08-06
fix Meta-Enter

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/340.

### DIFF
--- a/lib/runtime/console2.js
+++ b/lib/runtime/console2.js
@@ -61,12 +61,7 @@ export function activate (_ink) {
     client.boot()
   }
 
-  terminal.attachCustomKeyEventHandler(e => {
-    if (whitelistedKeybindingsREPL.indexOf(atom.keymaps.keystrokeForKeyboardEvent(e)) > -1) {
-      return false
-    }
-    return e
-  })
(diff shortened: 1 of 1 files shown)


lib/runtime/console2.js
```diff
@@ -257,8 +252,14 @@ function addLinkHandler (terminal) {
...
+  // Meta-Enter doesn't work properly with xterm.js atm, so we send the right escape sequence ourselves:
+  if (e.keyCode === 13 && (e.altKey || e.metaKey)) {
+    if (term.ty) {
+      term.ty.write('\x1b\x0d')
+    }
+    return false
+  } else if (binds.indexOf(atom.keymaps.keystrokeForKeyboardEvent(e)) > -1) {
     return false
   }
   return e
```
NHDaly: As you have it currently, the comment looks like it's describing the entire `if-elseif-end` block. Maybe you should put the comment inside the if? And/or also add a comment on the `else if` line?

pfitzseb: Fair point. Fixed.

PR #592, opened 2019-08-18
consistent types

aviatesk:
## purpose

refer to https://github.com/JunoLab/Atom.jl/pull/153 for details

## misc

refactors included
- allow preceding whitespaces in function calls: closes #588 (dup)
- ~~remove PlatformIO integration~~
- update menus / better messages using Markdown strings

### DIFF
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -175,7 +175,7 @@ module.exports =
   connectedError: (action = 'do that') ->
     if @isActive()
       atom.notifications.addError "Can't #{action} with a Julia client running.",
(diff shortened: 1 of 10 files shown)


aviatesk:
@pfitzseb maybe can I fire the old console.coffee stuff ? If we go with removing Platformio integration, there is no part that refers to the console.coffee anymore.

aviatesk:
> @pfitzseb maybe can I fire the old console.coffee stuff ?
> ~If we go with removing Platformio integration, there is no part that refers to the console.coffee anymore.~ nvm, it's not related at all.

It would make more sense to do this as a separate pr, thus just reverted 👍

PR #595, opened 2019-08-21
remove stuff

aviatesk:


pfitzseb:
Looks good. Probably also makes sense to remove the old console implementation from ink.

aviatesk:
yep, i already have that on my machine.
Will make a pr once I get on the Internet.

aviatesk:
xref: https://github.com/JunoLab/atom-ink/pull/214

PR #596, opened 2019-08-22
Adding useful toolbar buttons

aminya:
* Open Folder - useful for opening a project/package
* Select Julia working directory
* Auto Indent Selection
* Fold all - Unfold all: useful for code overview
* Show documentation of selection
* Markdown preview: for previewing Readme.md
* Open Settings
* Fullscreen
* Command Palette

image

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -382,7 +382,7 @@ config =
         title: 'Enable Extra Toolbar Buttons'
         type: 'boolean'
         default: true
-        description: 'Adds some useful toolbar buttons'
+        description: 'Adds some useful toolbar buttons (requires restart).'
         order: 11
   consoleOptions:
     type: 'object'
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -378,6 +378,12 @@ config =
             type: 'boolean'
             default: true
             order: 10
+      enableExtraToolbarButtons:
+        title: 'Enable Extra Toolbar Buttons'
+        type: 'boolean'
+        default: true
+        description: 'Adds some useful toolbar buttons'
+        order: 11
   consoleOptions:
     type: 'object'
     order: 4
--- a/lib/package/toolbar.coffee
+++ b/lib/package/toolbar.coffee
@@ -26,15 +26,17 @@ module.exports =
       tooltip: 'Open File...'
       callback: 'application:open-file'
 
-    @bar.addButton
-      icon: 'file-submodule'
-      tooltip: 'Open Folder...'
-      callback: 'application:open-folder'
+    if atom.config.get 'julia-client.uiOptions.enableExtraToolbarButtons'
+        @bar.addButton
+          icon: 'file-submodule'
+          tooltip: 'Open Folder...'
+          callback: 'application:open-folder'
 
-    @bar.addButton
-      icon: 'file-symlink-directory'
-      tooltip: 'Select Working Directory...'
-      callback: 'julia-client:select-working-folder'
+    if atom.config.get 'julia-client.uiOptions.enableExtraToolbarButtons'
+        @bar.addButton
+          icon: 'file-symlink-directory'
+          tooltip: 'Select Working Directory...'
+          callback: 'julia-client:select-working-folder'
 
     # Julia process
 
@@ -88,23 +90,26 @@ module.exports =
       tooltip: 'Format Code'
       callback: 'julia-client:format-code'
 
-    @bar.addButton
-      icon: 'indent'
-      callback: 'editor:auto-indent'
-      tooltip: 'Auto indent (selection)'
-      iconset: 'fa'
+    if atom.config.get 'julia-client.uiOptions.enableExtraToolbarButtons'
+        @bar.addButton
+          icon: 'indent'
+          callback: 'editor:auto-indent'
+          tooltip: 'Auto indent (selection)'
+          iconset: 'fa'
 
-    @bar.addButton
-      icon: 'chevron-right'
-      callback: 'editor:fold-all'
-      tooltip: 'Fold all'
-      iconset: 'fa'
+    if atom.config.get 'julia-client.uiOptions.enableExtraToolbarButtons'
+        @bar.addButton
+          icon: 'chevron-right'
+          callback: 'editor:fold-all'
+          tooltip: 'Fold all'
+          iconset: 'fa'
 
-    @bar.addButton
-      icon: 'chevron-down'
-      callback: 'editor:unfold-all'
-      tooltip: 'Unfold all'
-      iconset: 'fa'
+    if atom.config.get 'julia-client.uiOptions.enableExtraToolbarButtons'
+        @bar.addButton
+          icon: 'chevron-down'
+          callback: 'editor:unfold-all'
+          tooltip: 'Unfold all'
+          iconset: 'fa'
 
 
     # Windows & Panes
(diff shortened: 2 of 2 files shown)


aminya:
Works best if the toolbar is on bottom or top of the screen. I couldn't find any specific place to adjust this setting.

aviatesk:
Hi, 
I wonder those icons are really necessary for general Juno users. IMHO the added commands aren't strongly related to Juno, and should be extended by each user IMHO.

For example, actually each user can set tool bar icons according to your preference in your init script like:
> init.js
```js
const juliaClientWatcher = atom.packages.onDidActivatePackage((pkg) => {
  if (pkg.name !== 'julia-client') return;

  // Dispose Hydrogen keybindings for Julia files
  hydrogenKeybindsForJulia.dispose();

  // Create custom Julia-Client commands
  atom.commands.add('atom-workspace', {
    // Restart Julia Process
    'julia-client:restart-julia': async () => {
      const element = atom.workspace.getElement();
      if (!element) return;
      await atom.commands.dispatch(element, 'julia-client:kill-julia');
      atom.commands.dispatch(element, 'julia-client:start-julia');
    },
  });

  // Mimic Hydrogen's keybindings
  const juliaClientCommands = {
    'ctrl-c ctrl-c': 'julia-client:run-block',
    'ctrl-c ctrl-n': 'julia-client:run-and-move',
    'ctrl-c ctrl-d': 'julia-client:run-cell',
    'ctrl-c ctrl-b': 'julia-client:run-cell-and-move',
    'ctrl-c ctrl-r': 'julia-client:run-all',
    'ctrl-enter': 'julia-client:run-block',
    'shift-enter': 'julia-client:run-and-move',
    'alt-ctrl-enter': 'julia-client:run-cell',
    'alt-shift-enter': 'julia-client:run-cell-and-move',
    'ctrl-alt-shift-enter': 'julia-client:run-all',
    'alt-i': 'julia-client:show-documentation',
    'ctrl-c space': 'inline:clear-current',
    'ctrl-c ctrl-space': 'inline-results:clear-all',
  };
  const juliaClientMarkCommands = {
    'ctrl-c ctrl-d': 'julia-client:run-block',
    'ctrl-c ctrl-b': 'julia-client:run-and-move',
    'alt-ctrl-enter': 'julia-client:run-block',
    'alt-shift-enter': 'julia-client:run-and-move',
  };
  // Attach Julia-Client commands to .jl/.jmd files
  atom.keymaps.add(
    'init.js', {
      'atom-workspace': { 'alt-j ctrl-r': 'julia-client:restart-julia' },
      'atom-text-editor[data-grammar=\'source julia\'].emacs-plus:not([mini])': juliaClientCommands,
      'atom-text-editor[data-grammar=\'source weave md\'].emacs-plus:not([mini])': juliaClientCommands,
      'atom-text-editor[data-grammar=\'source julia\'].emacs-plus.mark-mode:not([mini])': juliaClientMarkCommands,
      'atom-text-editor[data-grammar=\'source weave md\'].emacs-plus.mark-mode:not([mini])': juliaClientMarkCommands,
    },
    1,
  );

  // Tweak Tool-Bar integration if activated
  if (atom.packages.isPackageLoaded('tool-bar')) {
    const toolBar = atom.packages.getLoadedPackage('tool-bar');
    if (!toolBar) {
      // eslint-disable-next-line no-console
      console.warning('Julia-Client: Failed to customize Tool-Bar integration');
      return;
    }
    const tb = toolBar.mainModule.provideToolBar()('julia-client');

    // Process
    tb.addButton({
      icon: 'flame',
      tooltip: 'Start Julia Process',
      callback: 'julia-client:start-julia',
    });
    tb.addButton({
      icon: 'globe',
      tooltip: 'Start Remote Julia Process',
      callback: 'julia-client:start-remote-julia-process',
    });
    tb.addButton({
      icon: 'zap',
      tooltip: 'Interrupt Julia Process',
      callback: 'julia-client:interrupt-julia',
    });
    tb.addButton({
      icon: 'sync',
      tooltip: 'Restart Julia Process',
      callback: 'julia-client:restart-julia',
    });
    tb.addButton({
      icon: 'trashcan',
      tooltip: 'Kill Julia Process',
      callback: 'julia-client:kill-julia',
    });

    tb.addSpacer();

    // In Process
    tb.addButton({
      icon: 'play',
      iconset: 'ion',
      tooltip: 'Run A Whole File',
      callback: 'julia-client:run-all',
    });
    tb.addButton({
      icon: 'format-float-none',
      iconset: 'mdi',
      tooltip: 'Format Code',
      callback: 'julia-client:format-code',
    });
    tb.addButton({
      icon: 'file-directory',
      tooltip: 'Set Working Directory',
      callback: 'julia-client:select-working-folder',
    });
    tb.addButton({
      icon: 'package',
      tooltip: 'Set Working Module',
      callback: 'julia-client:set-working-module',
    });

    tb.addSpacer();

    // Panes
    tb.addButton({
      icon: 'terminal',
      tooltip: 'Open Console Pane',
      callback: 'julia-client:open-console',
    });
    tb.addButton({
      icon: 'book',
      tooltip: 'Open Workspace Pane',
      callback: 'julia-client:open-workspace',
    });
    tb.addButton({
      icon: 'info',
      tooltip: 'Open Documentation Pane',
      callback: 'julia-client:open-documentation-browser',
    });
    tb.addButton({
      icon: 'graph',
      tooltip: 'Open Plot Pane',
      callback: 'julia-client:open-plot-pane',
    });
    tb.addButton({
      icon: 'bug',
      tooltip: 'Open Debugger Pane',
      callback: 'julia-debug:open-debugger-pane',
    });

    tb.addSpacer();

    // Meta
    tb.addButton({
      icon: 'gear',
      tooltip: 'Open Settings',
      callback: 'julia-client:settings',
    });
  }

  // Dispose this package watcher
  juliaClientWatcher.dispose();
});
```

aminya:
> I wonder those icons are really necessary for general Juno users. IMHO the added commands aren't strongly related to Juno, and should be extended by each user IMHO.

I disagree about letting the user add these. If they are expert enough they can remove them if they don't want them.

aminya:
I added the option for disabling extra buttons. So now if the user doesn't want these there is a simple checkmark to disable. Needs atom restart to see the result.

aminya:
Sorry, I had forgotten to commit `config.coffee`.

aminya:
@aviatesk could you please review this?

aviatesk:
Still I think these icons should be handled by the other packages or each user, since it's not related to Juno or not such core operations.
Each package (including Juno) provides icons and user selects them. If they don't like it, they can just customise according to their preferance. This is how the [tool-bar package](https://github.com/suda/tool-bar) exists and why I told you how to customise it in your init script.

Of course you can forcibly tweak it via configs like above, but still it need to change an users's tool-bar settings (like [this part of uber-juno installation](https://github.com/JunoLab/uber-juno/blob/master/lib/uber-juno.coffee#L20)) and then you would end up destroying each user's original tool-bar settings, i.e.: this kind of thing shouldn't happen so often IMHO.
And we currently have a lot of configs thus want to keep it simpler as possible.

aviatesk:
But maybe adding an icon to call `julia-client: settings` would be preferable since lots of user would see settings rather often.

Maybe @pfitzseb has some other opinion.

aminya:
I used the settings menu of Julia Client. I just added a simple checkmark.
image
I can improve the speed a little because I called to settings several time instead of defining a variable.

aviatesk:
I know, but please understand what https://github.com/JunoLab/atom-julia-client/pull/596#issuecomment-524607421 says

aminya:
I don't see this as destructive because the setting can be changed easily? What about setting the default value to `false`? 
Also, if we use `observe` instead of `get`, it won't need restarting.

aviatesk:
As you said in the first comment, these icons don't fit in the Juno's default tool-bar setting, thus we need to change it when user uses the config. This can be done in observing config value.

aminya:
The same way layout and position for other panes are adjusted toolbar position can be added to settings (?).
Personally, I don't think people will get mad if an option is added. Worst case scenario, in Julia Pro which is exclusively for Julia, this can be totally implemented.

aviatesk:
Yeah, but the biggest thing is that "no ones knows what icons should be default", and thus the icons exported by Juno shouldn't be so fat/complicated.
After all, using the default (simple) icons or customising via packages or init script is the way to user tool-bar.

pfitzseb:
Yeah, most of this should be a separate package (which definitely could be installed automatically via uber-juno).

> Open Folder - useful for opening a project/package

`Add Project Folder` seems more useful, generally.

> Select Julia working directory

Ideally we'd hook into the project view and add an entry to the context menu for this.

>  Auto Indent Selection

This doesn't seem to be super useful, considering we have auto formatting now.

> Fold all - Unfold all: useful for code overview

Might be useful.

> Show documentation of selection

This already is in the context menu, which is a better place for it imo.

> Markdown preview: for previewing Readme.md

This should be added to `markdown-preview`.

> Open Settings
    Fullscreen
    Command Palette

Hm, not sure if those are really all that useful, considering they also are in the application menu.

---

So in summary, the only thing I'd kinda like to see here would be `Julia: Open Package As Project Folder` and/or `Application: Add Project Folder`. Ideally we wouldn't add new buttons for those, but IIRC tool-bar doesn't provide nested menus and stuff like that.

aminya:
> Yeah, most of this should be a separate package (which definitely could be installed automatically via uber-juno).

If you think this makes things more organized I'm down for it. Maybe something like `Juno-Toolbar-plus`

aviatesk:
Going to close this since the discussion seems to be settled :)

aminya:
Haven't had much time recently, but I will create a separate package for toolbar buttons.

PR #601, opened 2019-09-02
add uri-handler for completions

aviatesk:
xref: https://github.com/JunoLab/Atom.jl/pull/164

### DIFF
--- a/lib/runtime/completions.js
+++ b/lib/runtime/completions.js
@@ -77,6 +77,9 @@ class AutoCompleteProvider {
     completion.replacementPrefix = completion._prefix ? completion._prefix : prefix
     const icon = this.makeIcon(completion.icon)
     if (icon) completion.iconHTML = icon
+    // the workaround for https://github.com/atom/autocomplete-plus/issues/868, but can appear verbose
+    if (!completion.description && completion.descriptionMoreURL)
+      completion.description = 'Click the link below to show docs:'
     return completion
   }
 
--- a/lib/julia-client.coffee
+++ b/lib/julia-client.coffee
@@ -63,3 +63,5 @@ module.exports = JuliaClient =
   provideAutoComplete: -> @runtime.provideAutoComplete()
 
   provideHyperclick: -> @runtime.provideHyperclick()
+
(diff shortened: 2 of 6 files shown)


lib/runtime/completions.js
```diff
@@ -77,6 +77,9 @@ class AutoCompleteProvider {
     completion.replacementPrefix = completion._prefix ? completion._prefix : prefix
     const icon = this.makeIcon(completion.icon)
     if (icon) completion.iconHTML = icon
+    // the workaround for https://github.com/atom/autocomplete-plus/issues/868, but can appear verbose
+    if (!completion.description && completion.descriptionMoreURL)
+      completion.description = 'Click the link below to show docs:'
```
aviatesk: One thing annoying around adding this feature is this:
without `description`, `descriptionMoreURL` won't work, and the workaround can appear verbose IMO.

aviatesk: this means:
image

If you feel annoying this, I can close this PR.
Can I hear your opinion, @pfitzseb ?

pfitzseb: Would
```
       completion.description = ' '
```
look too weird?

aviatesk: Yeah, autocomplete-plus package inserts an newline after `description` anyways...

pfitzseb: Then let's just fix that with 
```
diff --git a/styles/ink.less b/styles/ink.less
index fa65ab5..e5ab2f8 100644
--- a/styles/ink.less
+++ b/styles/ink.less
@@ -8,8 +8,13 @@
 @import "syntax-variables";
 
 autocomplete-suggestion-list.select-list.popover-list {
-  .suggestion-description-content {
-    overflow-y: hidden;
+  .suggestion-description {
+    line-height: 1;
+    .suggestion-description-content {
+      overflow-y: hidden;
+      display: inline;
+      white-space: normal;
+    }
   }
 }
```
image
image

aviatesk:
Sorry, I missed to push to master ...

lib/runtime/completions.js
aviatesk: Looks really good. Thanks for the fix :)

PR #606, opened 2019-09-03
layouts stuff for outline pane

aviatesk:


pfitzseb:
Oops. Yeah, we need those.

PR #607, opened 2019-09-04
warn when incompatible ink version detected

pfitzseb:


lib/julia-client.coffee
```diff
@@ -28,7 +32,16 @@ module.exports = JuliaClient =
...
   requireInk: (fn) ->
-    if atom.packages.isPackageLoaded "ink" then fn()
+    if atom.packages.isPackageLoaded("ink")
+      inkVersion = atom.packages.loadedPackages["ink"].metadata.version
+      if not atom.devMode and not semver.satisfies(inkVersion, INK_VERSION_COMPAT)
+        atom.notifications.addWarning "Potentially incompatible `ink` version detected.",
+          description: 'Please make sure to upgrade `ink` to a version compatible with `' +
+                        INK_VERSION_COMPAT + '`. Current version is `' + inkVersion + '`. ' +
+                        'If this is not the latest version, then install it manually with e.g. ' +
+                        '`apm install ink@x.y.z`.'
```
aviatesk:
```coffee
description:
  """
  Please make sure to upgrade `ink` to a version compatible with `#{INK_VERSION_COMPAT}`.
  Current version is `#{inkVersion}`. If this is not the latest version, then install
  it manuall with e.g. `apm install ink@x.y.z`.
  """
```
would look better imho tho.

pfitzseb: Ah, seems like I'm slowly forgetting all those fancy coffeescript features...

aviatesk: haha, they are no longer worth to remember😅

aviatesk:
The CI log seems to say that we should update `package-lock.json` as well ?

PR #613, opened 2019-09-13
Keybindings for all debugger commands

milesfrain:
### Existing bindings:
```
f10 step-to-next-expression
f11 step-into-function
shift+f11 finish-function
alt-x toggle-breakpoint
alt+shift+x toggle-conditional-breakpoint
```

### Proposed bindings:
```
f5 start-debugging
shift-f5 stop-debugging
f8 continue
shift-f8 step-to-selected-line
f9 toggle-breakpoint
shift-f9 toggle-conditional-breakpoint
f10 step-to-next-expression
shift-f10 step-to-next-line
f11 step-into-function
shift-f11 finish-function
```

The proposed bindings are closer to what users might expect from other IDEs. Using the [default VSCode bindings](https://code.visualstudio.com/docs/getstarted/keybindings#_debug) for inspiration.
### Differences from the VSCode defaults:
```
continue: change, f5 => f8
step-to-selected-line: new, shift-f8
toggle-conditional-breakpoint: new, shift-f9
step-to-next-expression: new, f10
step-to-next-line: change, f10 => shift-f10
```

### Unbound commands:
* `clear-all-breakpoints`
    * Could use a modified f9 command, but I'm concerned that users might confuse this for the toggle-conditional-breakpoint binding, and it would be really annoying to accidentally lose all your breakpoints.
* `open-debugger-pane`
    * Doesn't seem like this needs a hotkey.

#### Replacements:
```
toggle-breakpoint: alt-x => f9
toggle-conditional-breakpoint: alt-shift-x => shift-f9
```
#### Additions:
```
f5 start-debugging (see notes on this "new command" below)
shift-f5 stop-debugging
f8 continue
shift-f8 step-to-selected-line
shift-f10 step-to-next-line
```
#### New command:
* `start-debugging`.
    * This is not very important, and not implemented in this change. One possibility is to remember the last `Juno.@enter` or `Juno.@run` command.

### Note on line vs expression stepping:
Debating whether to use f10 for step-to-next-line and shift-f10 for step-to-next-expression or the other way around.
Motivation for line as f10 is to match other IDEs, even though VSCode doesn't have expression stepping.
Motivation for expression as f10 is to match the pattern established by f11 step-into-function and shift-f11 finish-function, where modifying with shift takes a bigger step. This is also one-fewer Juno binding change.

```
# keymap for testing these changes without incorporating pull request
# ~/.atom/keymap.cson
'atom-text-editor[data-grammar="source julia"]':
  #'f5': 'julia-debug:start-debugging' # Todo: add start-debugging command
  'shift-f5': 'julia-debug:stop-debugging'
  'f8': 'julia-debug:continue'
  'shift-f8': 'julia-debug:step-to-selected-line'
  'f9': 'julia-debug:toggle-breakpoint' # changed from alt-x
  'shift-f9': 'julia-debug:toggle-conditional-breakpoint' # changed from alt-shift-x
  #'f10': 'julia-debug:step-to-next-expression' # existing
  'shift-f10': 'julia-debug:step-to-next-line'
  #'f11': 'julia-debug:step-into-function' # existing
  #'shift-f11': 'julia-debug:finish-function' # existing
```

### DIFF
--- a/keymaps/julia-client.cson.cmd
+++ b/keymaps/julia-client.cson.cmd
@@ -8,7 +8,13 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
 ## Common
 # Debug operations
 'atom-text-editor[data-grammar="source julia"]':
+  'shift-f5': 'julia-debug:stop-debugging'
+  'f8': 'julia-debug:continue'
+  'shift-f8': 'julia-debug:step-to-selected-line'
+  'f9': 'julia-debug:toggle-breakpoint'
+  'shift-f9': 'julia-debug:toggle-conditional-breakpoint'
   'f10': 'julia-debug:step-to-next-expression'
+  'shift-f10': 'julia-debug:step-to-next-line'
   'f11': 'julia-debug:step-into-function'
   'shift-f11': 'julia-debug:finish-function'
 
@@ -27,8 +33,6 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
   'cmd-j cmd-d': 'julia-client:show-documentation'
   'cmd-j cmd-m': 'julia-client:set-working-module'
   'cmd-j cmd-f': 'julia-client:format-code'
-  'ctrl-x': 'julia-debug:toggle-breakpoint'
-  'ctrl-shift-x': 'julia-debug:toggle-conditional-breakpoint'
 
 # Julia REPL
 '.platform-darwin .julia-terminal':
--- a/keymaps/julia-client.cson.ctrl
+++ b/keymaps/julia-client.cson.ctrl
@@ -8,7 +8,13 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
 ## Common
 # Debug operations
 'atom-text-editor[data-grammar="source julia"]':
+  'shift-f5': 'julia-debug:stop-debugging'
+  'f8': 'julia-debug:continue'
+  'shift-f8': 'julia-debug:step-to-selected-line'
+  'f9': 'julia-debug:toggle-breakpoint'
+  'shift-f9': 'julia-debug:toggle-conditional-breakpoint'
   'f10': 'julia-debug:step-to-next-expression'
+  'shift-f10': 'julia-debug:step-to-next-line'
   'f11': 'julia-debug:step-into-function'
   'shift-f11': 'julia-debug:finish-function'
 
@@ -28,8 +34,6 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
   'ctrl-j ctrl-m': 'julia-client:set-working-module'
   'ctrl-j ctrl-f': 'julia-client:format-code'
   'ctrl-shift-c': 'julia-client:interrupt-julia'
-  'alt-x': 'julia-debug:toggle-breakpoint'
-  'alt-shift-x': 'julia-debug:toggle-conditional-breakpoint'
 
 # Julia REPL
 '.platform-win32 .julia-terminal,


pfitzseb:
Nice, it's great to have a fairly consistent keymap for debugging!

I've opened a PR that shows the correct keybindings in the toolbar. The missing keybindings in the various menus are a limitation of how Atom autodetects them, I think -- we've worked around that for the toolbar, but I don't think we can do the same for menus.

One complication I just noticed is that `F11` is already resolved as `window:toggle-full-screen`, so we should either add a FAQ entry about that or supply an alternative keybindings.

milesfrain:
#### Here's a summary of `F11` behaviors:
* Body of .jl file active. Assuming that's what `atom-text-editor[data-grammar="source julia"]` means.
    * `julia-debug:step-into-function`
* Body of REPL/Terminal active
    * **Nothing ?!**
* Any other Atom element selected. Including REPL tab and .jl file tabs.
    * `window:toggle-full-screen`

Here are my thoughts on what should happen next:
* Map `F11` to `debug:step-into` only when a debug session is active.
    * This requires replacing `atom-text-editor[data-grammar="source julia"]` with something that represents debugging state.
    * Other editors [use F11 for both fullscreen](https://code.visualstudio.com/docs/getstarted/keybindings#_display) and debug:step, so that would be ideal for Juno too.
* Let `F11` also activate fullscreen (or debug:step) when REPL is in focus.
    * A simple workaround is adding an `F11` entry to the terminal/REPL mapping section, but that suggests that all Juno keymappings need to be duplicated here to avoid being absorbed by the terminal. Hopefully there's an option to let these keys pass through.
https://github.com/JunoLab/atom-julia-client/blob/082d338f717079e44699d17f23828a61f15aa920/keymaps/julia-client.cson.ctrl#L38-L44
* Use uniform naming for `Terminal`, `REPL`, and `Console`.
    * `Juno > Open Console` opens an Atom tab titled `REPL`
    * `Juno > Open REPL` opens an external `Terminal` window.
    * I think the following should be used instead:
        * `Juno > Open REPL` opens an Atom tab titled `REPL`
        * `Juno > Open Terminal` opens an external `Terminal` window.

pfitzseb:
> A simple workaround is adding an F11 entry to the terminal/REPL mapping section, but that suggests that all Juno keymappings need to be duplicated here to avoid being absorbed by the terminal. Hopefully there's an option to let these keys pass through.

There isn't -- a terminal application might want to use those keybindings as well, so we can't just pass them through all the time. Adding `F5`...`F11` to the [`whitelistedKeybindingsREPL`](https://github.com/JunoLab/atom-julia-client/blob/7f83d0696942bf838e6c007b2c591b0d07a9123f/lib/package/config.coffee#L436) config entry as well as the normal Atom keybindings is the best solution here.
Having the `Fx` keybindings consistent in Julia editors and the REPL would be great.

Would also be much appreciated if you could uniformize the Terminal/REPL/Console terminology, but we can also postpone that for another PR (and it doesn't need to hold up this one).

aviatesk:
> Would also be much appreciated if you could uniformize the Terminal/REPL/Console terminology, but we can also postpone that for another PR (and it doesn't need to hold up this one).

This would be really great, but I think it gonna be such a big/destructive refactors that is better to be done within the separate PR imho :)

milesfrain:
I tried different combinations of adding the `F5`...`F11` keys to the whitelist and even moving the debug commands to different sections of the keymap file (`atom-workspace` and `julia-terminal`, but none of these combinations have allowed the debugger keys to work when the REPL is in focus.

pfitzseb:
Closing in favor of #619. Thanks for getting the ball rolling on this, @milesfrain!

PR #618, opened 2019-09-18
file jumps

aviatesk:
This feature won't fit well into the naming `julia-client: goto-symbol`, but I implemented this under the command as well anyway.

### DIFF
--- a/lib/runtime/goto.js
+++ b/lib/runtime/goto.js
@@ -1,15 +1,42 @@
 /** @babel */
 
+import path from 'path'
+import { Point, Range } from 'atom'
+
 import { client } from '../connection'
 import modules from './modules'
 import { isValidScopeToInspect } from '../misc/scopes'
-import { getWord, isValidWordToInspect, wordRegex } from '../misc/words'
+import { getWord, isValidWordToInspect } from '../misc/words'
 
 const methods = client.import('methods')
+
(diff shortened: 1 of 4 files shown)


aviatesk:
I started feeling like our `hyperclick` features are now so improved, and it's better to be added to the uber-juno deps.

lib/runtime/goto.js
```diff
@@ -1,17 +1,44 @@
...
+
+    if (!match) return false
+
+    match = text.match(filePathRegex)
+    const filePathBody = match[0].replace(/"/g, '')
+    const dirPath = path.dirname(editor.getPath())
+    const filePath = path.join(dirPath, filePathBody)
+    const columnStart = match.index + 1
+    const columnEnd = columnStart + filePathBody.length
+    if (columnStart <= column && column <= columnEnd) {
```
pfitzseb: I think we need to check if `filePath` actually is a file somewhere, otherwise you'll end up in an empty editor, which might be confusing.

aviatesk: I intentionally didn't add the check, assuming it would be also used for file creation, but does this sound weird to you ?

aviatesk: Umm yeah, creating a new file is definitely not a `goto/jump` stuff.
Will address this.

aviatesk: f7afd54 :)

PR #619, opened 2019-09-19
Debug keymap

pfitzseb:
Supersedes https://github.com/JunoLab/atom-julia-client/pull/613. Don't forget to call `apm install` in the `julia-client` dir to update your actual keymap.


lib/runtime/debugger.js
```diff
@@ -67,17 +67,28 @@ export function close () {
...
   if (!active) {
-    atom.notifications.addError('You need to be debugging to do that.', {
-      description: 'You can start debugging by calling `Juno.@enter f(args...)` from the integrated REPL.',
-      dismissable: true
-    })
+    // Only show an error when toolbar button or command is used directly. `ev.originalEvent` is
+    // a `KeyboardEvent` if this was triggered by a keystroke.
+    if (ev.originalEvent === undefined) {
+      atom.notifications.addError('You need to be debugging to do that.', {
+        detail: 'You can start debugging by calling `Juno.@enter f(args...)` from the integrated REPL.',
```
aviatesk: Really small stuff, but I believe `description` would be better here, since it can render the code block as it is

pfitzseb: Oh yeah, of course. I just messed up the rebase ;)

PR #620, opened 2019-09-19
exclude default completion provider

pfitzseb:
Works pretty well for me -- snippets are still there (they have `inclusionPriority: 1`), as are latex-completions (with `inclusionPriority: 5`).

The one thing we need to think about is removing the `disableForSelector = '.source.julia .comment'` line, because that means you still end up with the default completions in comments. Thoughts?

### DIFF
--- a/lib/runtime/completions.js
+++ b/lib/runtime/completions.js
@@ -18,7 +18,7 @@ const completions = client.import('completions')
 class AutoCompleteProvider {
   selector = '.source.julia'
   disableForSelector = `.source.julia .comment`
-  excludeLowerPriority = false
+  excludeLowerPriority = true
   inclusionPriority = 3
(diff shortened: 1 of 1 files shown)


aviatesk:
> The one thing we need to think about is removing the disableForSelector = '.source.julia .comment' line, because that means you still end up with the default completions in comments. Thoughts?

Honestly I want to keep the `disableForSelector` -- I found the default provider fits in the comment scope, since we can use arbitrary texts in comments and the naive completions from the current buffer _could_ be useful (at least for me).

pfitzseb:
Fair enough. What do you think about the actual PR? So far I haven't really missed anything, so imho we should go with this.

aviatesk:
Appears good. Merged :)

PR #622, opened 2019-09-26
updates:

aviatesk:
~~- move weave settings into julia-client~~
- more friendly config description

### DIFF
--- a/keymaps/julia-client.cson.cmd
+++ b/keymaps/julia-client.cson.cmd
@@ -5,7 +5,6 @@ register only in Julia-scoped places (e.g. Julia-syntax buffer, console)
 Any global commands should either be non-default or, ideally, prefixed with `C-J`.
 ###
 
-## Common
 # Debug operations
 '.platform-darwin atom-text-editor[data-grammar="source julia"]:not(.mini),
   ink-terminal.julia-terminal,
@@ -20,7 +19,6 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
   'f11': 'julia-debug:step-into-function'
   'shift-f11': 'julia-debug:finish-function'
 
-## macOS
 # Julia atom-text-editor
 '.platform-darwin atom-text-editor[data-grammar="source julia"]':
   'cmd-enter': 'julia-client:run-block'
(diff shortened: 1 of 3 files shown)


keymaps/julia-client.cson.cmd
```diff
@@ -36,6 +34,19 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
...
+ .platform-darwin atom-text-editor[data-grammar="source weave latex"]':
+  'cmd-enter': 'julia-client:run-block'
+  'shift-enter': 'julia-client:run-and-move'
+  'cmd-shift-enter': 'julia-client:run-all'
+  'alt-enter': 'julia-client:run-cell'
+  'alt-shift-enter': 'julia-client:run-cell-and-move'
+  'alt-down': 'julia-client:next-cell'
+  'alt-up': 'julia-client:prev-cell'
+  'cmd-j cmd-g': 'julia-client:goto-symbol'
+  'cmd-j cmd-d': 'julia-client:show-documentation'
```
pfitzseb: I feel like this belongs in language-weave -- what's the reason for moving it here?

aviatesk: Umm because julia-client defines those commands ?

I mean, when we change the name of the command or something, and accordingly change `keymaps/` or `menus/`, it would be easier to change that for weave files, if we have them here.
(Like avoiding the case like that the context-menu for julia-client is now "Juno" but language-weave one is still "Julia-Client")

pfitzseb: Hm. language-weave depends on julia-client, so they should make sure to keep up with upstream (and yes, that means I should've made that change there as well :P).

IMHO we should keep those packages as separate as possible.

aviatesk: > language-weave depends on julia-client, so they should make sure to keep up with upstream

Sounds fair. Then, let's go with the current direction :)

Do you mind allowing the write access to language-weave for members in "Commiters" team ?
Then I could make such a tiny changes easily.

pfitzseb: Alright, you should have write access now.

aviatesk: confirmed, thx ;)

aviatesk:
will rebase a bit

aviatesk:
will merge once I confirm CI passes

PR #624, opened 2019-09-27
enhanced symbols

aviatesk:
ref: https://github.com/JunoLab/Atom.jl/pull/192

### DIFF
--- a/lib/misc/blocks.js
+++ b/lib/misc/blocks.js
@@ -102,27 +102,22 @@ export function moveNext (ed, sel, range) {
   return sel.setBufferRange([[to, Infinity], [to, Infinity]])
 }
 
-function getRanges (ed, row) {
-  const ranges = row ?
-    [{
-      selection: undefined,
-      range: getRange(ed, row)
-    }] :
-    ed.getSelections().map(sel => {
-      return {
-        selection: sel,
-        range: sel.isEmpty() ?
-          getRange(ed, sel.getHeadBufferPosition().row) :
-          getSelection(ed, sel)
-      }
-    })
+function getRanges (ed) {
+  const ranges = ed.getSelections().map(sel => {
+    return {
+      selection: sel,
+      range: sel.isEmpty() ?
+        getRange(ed, sel.getHeadBufferPosition().row) :
+        getSelection(ed, sel)
+    }
+  })
   return ranges.filter(({ range }) => {
-    return range != null && ed.getTextInBufferRange(range).trim()
+    return range && ed.getTextInBufferRange(range).trim()
   })
 }
 
-export function get (ed, row) {
-  return getRanges(ed, row).map(({ range, selection }) => {
+export function get (ed) {
+  return getRanges(ed).map(({ range, selection }) => {
     return {
       range,
       selection,
@@ -133,10 +128,12 @@ export function get (ed, row) {
 }
 
 export function getLocalContext (editor, row) {
-  const block = get(editor, row)[0]
+  const range = getRange(editor, row)
+  const context = range ? editor.getTextInBufferRange(range) : ''
+  const startRow = range ? range[0][0] : null
   return {
-    context: block ? block.text : '',
-    startRow:  block ? block.line : 1
+    context,
+    startRow
   }
 }
 
--- a/lib/misc/blocks.js
+++ b/lib/misc/blocks.js
@@ -11,21 +11,21 @@ function getLine (ed, l) {
 }
 
 function isBlank ({line, scope}, allowDocstrings = false) {
-  for (let s of scope) {
+  for (const s of scope) {
     if (/\bcomment\b/.test(s) || (!allowDocstrings && /\bdocstring\b/.test(s))) {
       return true
     }
   }
   return line.match(/^\s*(#.*)?$/)
 }
-function isEnd ({line, scope}) {
+function isEnd ({ line }) {
   return line.match(/^(end\b|\)|\]|\})/)
 }
-function isCont ({line, scope}) {
+function isCont ({ line }) {
   return line.match(/^(else|elseif|catch|finally)\b/)
 }
 function isStart (lineInfo) {
-  for (let s of lineInfo.scope) {
+  for (const s of lineInfo.scope) {
     if (/\bstring\b/.test(s)) {
       return false
     }
@@ -84,7 +84,7 @@ function getSelection (ed, sel) {
 
 export function moveNext (ed, sel, range) {
   // Ensure enough room at the end of the buffer
-  let row = range[1][0]
+  const row = range[1][0]
   let last
   while ((last = ed.getLastBufferRow()) < (row+2)) {
     if ((last !== row) && !isBlank(getLine(ed, last))) {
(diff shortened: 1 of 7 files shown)


lib/ui/docs.js
```diff
@@ -4,8 +4,10 @@ import { client } from '../connection'
 import { CompositeDisposable } from 'atom'
 const views = require('./views')
 
-const { searchdocs, methods, moduleinfo, regenerateCache } =
-      client.import({rpc: ['searchdocs', 'methods', 'moduleinfo'], msg: ['regenerateCache']})
+const { searchdocs, goto, moduleinfo, regenerateCache } =
+      client.import({rpc: ['searchdocs', 'goto', 'moduleinfo'], msg: ['regenerateCache']})
+
+const grammar = atom.grammars.grammarForScopeName('source.julia')
```
pfitzseb: Same as above

lib/runtime/goto.js
```diff
@@ -2,20 +2,55 @@
...
 import { client } from '../connection'
 import modules from './modules'
 import { isValidScopeToInspect } from '../misc/scopes'
 import { getWord, getWordRangeAtBufferPosition, isValidWordToInspect } from '../misc/words'
+import { getLocalContext } from '../misc/blocks'
 
-const methods = client.import('methods')
+const goto = client.import('goto')
 
+const grammar = atom.grammars.grammarForScopeName('source.julia')
```
pfitzseb: This *needs* to be in `activate` or wherever it is actually used, otherwise it'll break if language-julia is loaded after julia-client (and that ordering isn't stable).

lib/misc/blocks.js
```diff
@@ -102,32 +102,48 @@ export function moveNext (ed, sel, range) {
...
       selection,
       line: range[0][0],
       text: ed.getTextInBufferRange(range)
-    })
+    }
+  })
+}
+
+export function getLocalContext (editor, row) {
+  const block = get(editor, row)[0]
```
pfitzseb: Seems like the `get(ed, row)`  and `getRanges(ed, row)` functionality is only used here, so maybe it'd be better to move that logic here and use `getRange` directly?

lib/runtime/goto.js
```diff
@@ -2,20 +2,55 @@
...
+      atom.workspace.observeTextEditors(editor => {
+        if (
+          !this.refreshFiles.has(editor.getPath()) &&
+          editor.getGrammar() === grammar
+        ) {
+          const buffer = editor.getBuffer()
+          const bufferSubscriptions = new CompositeDisposable()
+          bufferSubscriptions.add(buffer.onDidSave(({ path }) => {
+            if (!this.refreshFiles.has(path)) {
+              this.refreshFiles.add(path)
```
pfitzseb: You can just unconditionally `add(path)` here, no? `refreshFiles` is a `Set` after all, so adding twice doesn't hurt.

lib/runtime/goto.js
```diff
@@ -2,20 +2,55 @@
...
+            }
+          }))
+          bufferSubscriptions.add(buffer.onDidReload(() => {
+            const path = buffer.getPath()
+            if (!this.refreshFiles.has(path)) {
+              this.refreshFiles.add(path)
```
pfitzseb: As above.

lib/runtime/goto.js
```diff
@@ -2,20 +2,55 @@
...
 class Goto {
   activate (ink) {
     this.ink = ink
+    this.subscriptions = new CompositeDisposable()
+    this.refreshFiles = new Set()
+    this.subscriptions.add(
+      atom.workspace.observeTextEditors(editor => {
+        if (
+          !this.refreshFiles.has(editor.getPath()) &&
+          editor.getGrammar() === grammar
```
pfitzseb: I've used `ed.getGrammar().id === 'source.julia'` checks for this, might be worth making that change.

aviatesk:
> AFAICT we're only updating the symbol list on buffer save and reload, right? That's different from the outline, which is updated when the onDidStopChanging event is fired. It would be nice to be consistent here, but I'm not sure which way we should go.

Yeah, we should make it consistent.
I observed buffer save/reload, since then we don't need to path the buffer text via an atom-julia-client channel.

But now I think observing `onDidStopChanging` is better here, since I experienced cases when I want to have symbol jumps when I still don't want to save it.

lib/runtime/goto.js
aviatesk: Ah, I've forgotten that. Thanks for reminding me :)
Will fix soon.

lib/runtime/workspace.coffee
```diff
@@ -5,7 +5,9 @@
 
 modules = require './modules'
 
-{workspace, clearLazy} = client.import rpc: 'workspace', msg: 'clearLazy'
+{ workspace, goto, clearLazy } = client.import rpc: ['workspace', 'goto'], msg: 'clearLazy'
+
+grammar = atom.grammars.grammarForScopeName 'source.julia'
```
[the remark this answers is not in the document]
aviatesk: remove this line, since I will move to implement module-based symbol caching, and then this fallback path would complicate things.

lib/ui/docs.js
aviatesk: same as workspace.coffee

lib/runtime/goto.js
aviatesk: year, maybe I forget what `Set` is for at all xD

lib/runtime/goto.js
aviatesk: Okay, will use that instead.

pfitzseb:
>  I observed buffer save/reload, since then we don't need to path the buffer text via an atom-julia-client channel.

Yeah, I saw that. Let's keep the best of both worlds and `read` the file in Julia if the editor has no pending changes and send the buffer contents if it does.

aviatesk:
> Let's keep the best of both worlds and read the file in Julia if the editor has no pending changes and send the buffer contents if it does.

I could take this direction, but I thought just sending buffer text _always_ is going to be a lot simpler here, since sending a single buffer text per `goto` call won't cost so much. Do you think it would be worth to making two branches here ?

pfitzseb:
Not sure, to be honest. Lemme try to find a server to try this together with a remote Julia session. For local development sending the buffer content should be strictly better (but not by much).

aviatesk:
> Lemme try to find a server to try this together with a remote Julia session
Thanks this should help us greatly.

lib/misc/blocks.js
aviatesk: Actually `get(ed, row)` is used in [completions.js](https://github.com/JunoLab/atom-julia-client/blob/master/lib/runtime/completions.js#L59-L62), thus I refrained from migrating `get` and `getRanges` into `getLocalContext`.

But if you don't mind the small refactor in completions.js, I will do so :)

aviatesk: Ah also [evaluation.coffee](https://github.com/JunoLab/atom-julia-client/blob/master/lib/runtime/evaluation.coffee#L32) uses `get`.

So... I feel like I want to keep `get` and `getRanges` as is ?

pfitzseb: Mh? `completions.js` and `evaluation.coffee` only use `get(ed)`.  I don't care too much about this, so go with whatever you want ;)

aviatesk: yeah, you're right. I always get confused with function calls in JS ...

aviatesk: okay, maybe 8122f31 will simplify the logic

aviatesk:
So I'm coming up with 3 approaches to handle unsaved buffer changes:

1. passing a buffer text only if the active text editor has been modifed: the unsaved changes in non active buffers can get ignored
2. passing an active buffer text _always_: the unsaved changes in non active buffers can get ignored, simplest choice
3. tracking active (Julia) buffers, and passing all the modified ones (even though some of them may not be reached in the symbols search), and Julia will use the passed text if the symbol search lands to the file: can handle all the unsaved buffer changes

So 1 is the lowest latency option, and 3 is the heaviest, and 2 is the medium.

I'm not sure which way we should start with, but maybe starting to implement 3 and simplify the logic into 1 or 2 if we find performance pitfalls is reasonable.

Thoughts ?

pfitzseb:
We'd only ever need to pass the last modified buffer if we cache everything else in Julia, right? IMHO that'd be the best way to handle this.

aviatesk:
> We'd only ever need to pass the last modified buffer if we cache everything else in Julia, right? IMHO that'd be the best way to handle this.

No, I meant cases when we have multiple files open (so multiple panes open) that have unsaved changes. My point is whether we should handle all the unsaved buffers in the workspace, or just focus on the (very) active buffer.

aviatesk:
This is because I don't want to call `goto` every time we made changes, but just add the `TextEditor` instance to `this.refreshFiles` (and they will get refreshed within the next (explicit) `goto` calls).

aviatesk:
Um, but maybe calling `goto` everytime we've made changes to the buffer might be robust here.

Now I'm thinking the workflow below:
- `julia-client: goto-symbol` and symbol clicks within docpane/workspace will call `goto`, which will do module-based symbol search
  * use caches if they exist
  * for file searching, Revise approach first and fallback to static file parsing
- the new message `regeneratesymbols` will be called on `onDidStopedChanging`, and it updates symbols that can be searched from the passed buffer texts

pfitzseb:
Yep, that's precisely what I had in mind.

We'll obviously need a three-step cache (`Module` -> `File` -> `Symbol`), but that shouldn't be hard to implement.

aviatesk:
yeah, I'm almost there.

I wondered calling `regenerateSymbols` every time we modify a file can be annoying and occupy computation resources, but AFAICT, the performance of current implementation I have in my machine is not so bad.

pfitzseb:
If necessary we can debounce it more, but considering that it's usually local to the current editor we're looking at ~20ms at most on my machine (and the outline also already does the same thing, no?)-

aviatesk:
> and the outline also already does the same thing, no?

Yes, exactly, honestly I didn't thought of this fact at all, haha.
So this approach should definitely work well as far as we keep `toplevelitems` from being slow.

Thanks so much for your directions, as always :)

aviatesk:
> Would be good to not duplicate work here and generate the outline and symbols for a file in one go.

Addressed :)

PR #626, opened 2019-10-03
Update toolbar for missing icons

daviehh:
Some icons have gone missing following the most recent update to the toolbar package because the ion icons now need the `md-` prefix; also, `stop` is now called `square`.

see, e.g.

https://github.com/suda/tool-bar/releases/tag/v1.2.0

https://github.com/suda/tool-bar/issues/264


pfitzseb:
Thanks!

PR #636, opened 2019-10-09
add option to unset proxy env vars

### LINKED ISSUE #632: Add setting to not pass `proxy`/`https-proxy` to spawned Julia processes
Because libgit doesn't deal with that setting very well.

pfitzseb:
Fixes #632.


lib/package/config.coffee
```diff
@@ -510,6 +510,15 @@ config =
     type: 'array'
     default: ['source.julia', 'source.weave.md', 'source.weave.latex']
     order: 6
+  disableProxy:
```
aviatesk: This seems better to be under `Julia Options`, because all the other configs around Julia process is so ?

pfitzseb: It's not strictly for Julia though (and has an impact on all internal terminals, not just the REPL).

aviatesk: yep, just asked it for a case if this was only intended to fix the package-installation issue.

PR #641, opened 2019-10-23
rename refactor

aviatesk:
xref: https://github.com/JunoLab/Atom.jl/pull/202


aminya:
@aviatesk is this PR still relevant? Could you rebase so I know which files are under editing?

PR #643, opened 2019-10-24
don't update symbols on editor switching

aviatesk:
fixes https://github.com/JunoLab/Juno.jl/issues/407
ref: https://github.com/JunoLab/Atom.jl/pull/205

### DIFF
--- a/lib/runtime/outline.js
+++ b/lib/runtime/outline.js
@@ -18,7 +18,7 @@ export function activate (ink) {
   subs.add(client.onDetached(() => pane.setItems([])))
 
   let edSubs = new CompositeDisposable()
-  subs.add(atom.workspace.onDidChangeActiveTextEditor(ed => {
+  subs.add(atom.workspace.onDidChangeActiveTextEditor(throttle(ed => {
     if (!ed) return
 
     edSubs.dispose()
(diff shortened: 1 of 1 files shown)


lib/runtime/outline.js
```diff
@@ -18,7 +18,7 @@ export function activate (ink) {
   subs.add(client.onDetached(() => pane.setItems([])))
 
   let edSubs = new CompositeDisposable()
-  subs.add(atom.workspace.onDidChangeActiveTextEditor(ed => {
+  subs.add(atom.workspace.onDidChangeActiveTextEditor(throttle(ed => {
```
aviatesk: I think `throttle` here would be reasonable, and won't make Julia hard-work when consecutive editor switches.

pfitzseb: Yeah, good call.

lib/runtime/outline.js
```diff
@@ -60,14 +60,18 @@ export function activate (ink) {
       pane.setItems(outline)
     }, 300)))
 
-    updateEditor(ed).then(outlineItems => {
+    updateEditor(ed, {
+      noUpdateSymbols: true
```
pfitzseb: Nitpick, but I feel like `updateSymols: false` would be easier to parse (here and in the Julia code).

aviatesk:
good to go.

PR #648, opened 2019-10-28
follow646

aviatesk:
`currentDir` can now error when invoked _not_ from tree-view context menu.
I also found `cdProject` can error as well.

All the evil was the variable hoisting.

### DIFF
--- a/menus/julia-client.cson
+++ b/menus/julia-client.cson
@@ -1,7 +1,6 @@
 'context-menu':
   'atom-text-editor[data-grammar="source julia"]': [
     {type: 'separator'}
-
     {
       label: 'Juno',
       submenu: [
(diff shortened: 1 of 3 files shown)


menus/julia-client.cson
```diff
@@ -16,10 +15,17 @@
...
-
     {type: 'separator'}
   ]
+
   '.tree-view li.directory': [
-    { label: 'Juno: Work in Folder', command: 'julia-client:work-in-current-folder' }
-    { label: 'Juno: Activate Project', command: 'julia-client:activate-current-folder' }
+    {type: 'separator'}
+    {
+      label: 'Juno',
```
aviatesk: If you don't want nesting here, please let me know :)

pfitzseb:
Ugh, I thought I'd fix that. Thanks.

Issue #649, opened 2019-10-28
Bug: change to current file's folder menu item broken


daviehh:
Just noticed that the menu item `Juno > Working Directory > Current File's Folder` is not working: with some file `foo.jl` and REPL open, clicking that menu and `pwd()` in REPL does not change to `foo.jl`'s directory (default to ~); the command palette Julia Client: work in current folder works.

Looks like in this recent commit https://github.com/JunoLab/atom-julia-client/commit/fddc075a6bb7925c2b731ca7d104968be076866c,  `work-in-file-folder` is changed to `work-in-current-folder`, and the corresponding command pallet is updated in that commit but the menu file still uses `work-in-file-folder`:

https://github.com/JunoLab/atom-julia-client/blob/d9b50d748ccc78b6cee9d1625bd7c523b4649eff/lib/package/menu.coffee#L38

maybe that should also be changed to `work-in-current-folder`? Looks like `work-in-file-folder` is no longer used here: https://github.com/JunoLab/atom-julia-client/search?q=work-in-file-folder&unscoped_q=work-in-file-folder

PR #650, opened 2019-10-28
menu: current file's folder menu item fix

daviehh:
For https://github.com/JunoLab/atom-julia-client/issues/649


aviatesk:
thanks for the fix !

PR #651, opened 2019-10-29
RFC: strip trailing dots

aviatesk:


pfitzseb:
Sorry, I was kinda out of commission for the last couple of days.

Tbh, I don't like the 
```
getWordAndRange (editor, options = {
  bufferPosition: undefined,
  wordRegex: wordRegex,
  beginWordRegex: undefined,
  endWordRegex: undefined
})
``` 
API. Do you anticipate any need for user defined regexps there?
Imho it would be cleaner to just split the current "word" on `.`s and indicate in which the cursor is. Or maybe return a
```
[
  {
    text: 'Foo',
    range: [0, 2]
  },
  {
    text: 'bar',
    range: '[4, 6]
  }
]
```
for `Foo.bar`.

aviatesk:
> Sorry, I was kinda out of commission for the last couple of days.

No worries :)

> Tbh, I don't like the
> 
> ```
> getWordAndRange (editor, options = {
>   bufferPosition: undefined,
>   wordRegex: wordRegex,
>   beginWordRegex: undefined,
>   endWordRegex: undefined
> })
> ```
> 
> API. Do you anticipate any need for user defined regexps there?

yeah, fair enough. I made such a complex API since I thought it would be more extensible, but I'm not coming up with the use cases of the user-defined regexes other than this case for now, to be honest.

> Imho it would be cleaner to just split the current "word" on `.`s and indicate in which the cursor is. Or maybe return a
> 
> ```
> [
>   {
>     text: 'Foo',
>     range: [0, 2]
>   },
>   {
>     text: 'bar',
>     range: '[4, 6]
>   }
> ]
> ```
> 
> for `Foo.bar`.

And so this would be better solution -- stripping trailing dots would be fairly easy with this approach, too. I will try to change the function APIs then :)

aviatesk:
okay, e9e9845 hopefully simplifies the logic.

PR #654, opened 2019-11-04
compat with old julia versions

pfitzseb:


gcalajo:
hello thanks for the script. I am not pratic with Atoms where it should go this  code in order to make Julia v 6.4 working? Thanks in advance

PR #657, opened 2019-11-06
open terminal from a text editor's directory

aviatesk:
I often want this when editing files outside of a current project.

Or maybe creating commands like `julia-client: new-(remote-)terminal-from-active-directory` (which will just start a terminal within a currently active editor's directory) is more preferable ?

### DIFF
--- a/lib/package/menu.coffee
+++ b/lib/package/menu.coffee
@@ -44,7 +44,7 @@ module.exports =
       {
         label: 'Environment',
         submenu: [
-          {label: 'Current File\'s Folder', command: 'julia-client:activate-current-folder'}
+          {label: 'Environment in Current File\'s Folder', command: 'julia-client:activate-environment-in-current-folder'}
           {label: 'Default Environment', command: 'julia-client:activate-default-environment'}
         ]
       }
--- a/lib/package/menu.coffee
+++ b/lib/package/menu.coffee
@@ -36,9 +36,16 @@ module.exports =
         label: 'Working Directory'
         submenu: [
           {label: 'Current File\'s Folder', command: 'julia-client:work-in-current-folder'}
-          {label: 'Current Project\'s Folder', command: 'julia-client:work-in-project-folder'}
+          {label: 'Current Project\'s Folders', command: 'julia-client:work-in-project-folder'}
           {label: 'Home Folder', command: 'julia-client:work-in-home-folder'}
           {label: 'Select...', command: 'julia-client:select-working-folder'}
+        ]
+      }
+      {
+        label: 'Active Project',
+        submenu: [
+          {label: 'Current File\'s Folder', command: 'julia-client:activate-current-folder'}
+          {label: 'Default Environment', command: 'julia-client:activate-default-environment'}
         ]
       }
       {label: 'Select Working Module', command: 'julia-client:select-working-module'}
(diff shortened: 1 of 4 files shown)


pfitzseb:
Should probably model this after the `cd` and `activate-project` commands (either right click in tree view or open in current editors dir).

aviatesk:
> Should probably model this after the `cd` and `activate-project` commands (either right click in tree view or open in current editors dir).

So basically you agree with creating an new command ? Then we can call it from tree-view, editor menus, and the command itself with a text editor active.

aviatesk:
I found the current behaviour can be somewhat confusing

pfitzseb:
Yeah, we should have two commands. The current one is fairly useful, imho, especially since you can enter a path yourself, but something `cd`-like is probably more approachable.

aviatesk:
> The current one is fairly useful, imho, especially since you can enter a path yourself, but something `cd`-like is probably more approachable.

When we run a new command (`new-terminal-in-current-folder`, if I name it after `cd` and such), we usually just want to open a terminal from the very current directory (or selected one from tree-view), so the new command would be useful even without that, I guess.

aviatesk:
> I found the current behaviour can be somewhat confusing

and just for addition, I meant that the behaviour of this PR (including directories of workspace files in a selector list of `new-terminal`) can be confusing. The implementation so far is fairly good and not confusing imho as well :)

aviatesk:
Okay the `cd` and `activate-project` like command is implemented.
Menus and tree-view contexts are also updated.

lib/package/menu.coffee
```diff
@@ -36,11 +36,18 @@ module.exports =
         label: 'Working Directory'
         submenu: [
           {label: 'Current File\'s Folder', command: 'julia-client:work-in-current-folder'}
-          {label: 'Current Project\'s Folder', command: 'julia-client:work-in-project-folder'}
+          {label: 'Current Project\'s Folders', command: 'julia-client:work-in-project-folder'}
```
pfitzseb:
```suggestion
          {label: 'Select Project Folder', command: 'julia-client:work-in-project-folder'}
```
since this opens the project selector.

lib/package/menu.coffee
```diff
@@ -36,11 +36,18 @@ module.exports =
...
           {label: 'Home Folder', command: 'julia-client:work-in-home-folder'}
           {label: 'Select...', command: 'julia-client:select-working-folder'}
         ]
       }
+      {
+        label: 'Active Project',
```
pfitzseb: How about
```suggestion
        label: 'Environment',
```
?

lib/package/menu.coffee
```diff
@@ -66,7 +73,13 @@ module.exports =
 
       {type: 'separator'}
 
-      {label: 'New Terminal', command: 'julia-client:new-terminal'}
+      {
+        label: 'New Terminal'
+        submenu: [
+          {label: 'Current File\'s Folder', command: 'julia-client:new-terminal-from-current-folder'}
+          {label: 'Current Project\'s Folders', command: 'julia-client:new-terminal'}
```
pfitzseb:
```suggestion
          {label: 'Select Project Folder', command: 'julia-client:new-terminal'}
```

lib/package/menu.coffee
aviatesk: yeah sounds good.
As for this, how does `activate-environment-in-current-folder` sound to you instead of `activate-current-folder` ? IMO this sounds clearer (a bit verbose tho)

pfitzseb: Yeah, should be fine. It's not like anyone would need to type out those words ;)

PR #661, opened 2019-12-20
improve block finding algorithm yet again

pfitzseb:
Fixes https://discourse.julialang.org/t/strange-behavior-in-live-line-evaluation/32497 and makes a few general improvements.

### DIFF
--- a/lib/misc/blocks.js
+++ b/lib/misc/blocks.js
@@ -16,21 +16,21 @@ function isBlank ({line, scope}, allowDocstrings = false) {
       return true
     }
   }
-  return line.match(/^\s*(#.*)?$/)
+  return /^\s*(#.*)?$/.test(line)
 }
-function isEnd ({ line }) {
+function isEnd ({ line, scope }) {
+  for (const s of scope) {
+    if (/\bstring\.multiline\.end\b/.test(s)) {
+      return true
+    }
+  }
   return line.match(/^(end\b|\)|\]|\})/)
 }
 function isCont ({ line }) {
   return line.match(/^(else|elseif|catch|finally)\b/)
 }
 function isStart (lineInfo) {
-  for (const s of lineInfo.scope) {
-    if (/\bstring\b/.test(s)) {
-      return false
-    }
-  }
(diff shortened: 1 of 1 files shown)


aviatesk:
will try to take a look at this (I'm not sure I can understand that correctly)

lib/misc/blocks.js
```diff
@@ -46,13 +46,12 @@ function walkForward (ed, start) {
...
+
     if (isStart(lineInfo)) {
       break
     }
     if (isEnd(lineInfo)) {
-      if (!(forLines(ed, start, mark-1).length === 0)) {
-        end = mark
-      }
+      end = mark
     } else if (!(isBlank(lineInfo) || isStart(lineInfo))) {
```
aviatesk: nit: we can remove the second `isStart` condition since it's always `false` in this path.

lib/misc/blocks.js
```diff
@@ -16,21 +16,21 @@ function isBlank ({line, scope}, allowDocstrings = false) {
...
 }
 function isCont ({ line }) {
   return line.match(/^(else|elseif|catch|finally)\b/)
 }
 function isStart (lineInfo) {
-  for (const s of lineInfo.scope) {
-    if (/\bstring\b/.test(s)) {
-      return false
-    }
-  }
```
aviatesk: um, so if we remove this condition check, the code like below won't get selected as expected again:
```jl
ret = """
multiline string
...
""" |> somefunc
```

But I've not been able to figure out how to handle this while correctly run pieces like

```jl
str = "a line containing string scope"
str2 = "another line containing string scope"
...
```

pfitzseb: Urgh, true. I only tested
```
ret = """
  multiline string
  aksjdha
""" |> somefunc
```
Will take another look...

aviatesk: (I'm trying not to say let's use CSTParser for this too ...)

pfitzseb: Those cases work now.

aviatesk:
Just done manual tests (didn't look at code) but found this PR solves every case that is mentioned here !

pfitzseb:
Nice, let's go with this then. I'm sure someone will complain if I broke something else :P

PR #663, opened 2019-12-20
refactor completions.js:

aviatesk:
- use `blocks.getLocalContext` instead of `blocks.get`
- a bit rename refactor
- rm cacheCompletions


aviatesk:
CI failure seems to be caused by that the API changes (adding `prefix` in julia)

PR #668, opened 2020-01-13
register `julia:new-julia-file` command

aviatesk:


lib/package/commands.coffee
```diff
@@ -111,6 +111,12 @@ module.exports =
       'julia-debug:continue': (ev) => juno.runtime.debugger.continueForward(ev)
       'julia-debug:open-debugger-pane': => juno.runtime.debugger.open()
 
+      'julia:new-julia-file': =>
+        atom.workspace.open().then((ed) =>
+          gr = atom.grammars.grammarForScopeName('source.julia')
+          return ed.destroy() unless gr
```
pfitzseb: Why `destroy` the editor here?

aviatesk: Because then it's not "julia file" and so we want to show that this command failed ?
Notifying that `atom-language-julia` isn't installed in this path would be more friendly, though.

pfitzseb: Yeah, I feel like keeping the editor open and showing a warning or whatever is a bit nicer.

aviatesk: please change the notification wording as you want :)

pfitzseb: Actually, we're checking whether language-julia is installed on start [here](https://github.com/JunoLab/atom-julia-client/blob/master/lib/julia-client.coffee#L50-L59), I guess we can remove all checks here.

aviatesk: yeah, I borrowed the notification message from there, and felt like we are doing something boilerplate.

Let's just go with `ed.setGrammar(gr) if gr` then.

PR #672, opened 2020-01-24
update profiler

pfitzseb:


styles/julia-client.less
```diff
@@ -97,6 +97,25 @@ atom-text-editor.editor {
...
+        background: fade(@background-color-warning, 20%);
+      }
+    }
+    &.isgc > div {
+      border: solid fade(@background-color-error, 20%) 1px;
+      background: fade(@background-color-error, 10%);
+      &:hover {
+        background: fade(@background-color-error, 20%);
+      }
+    }
```
aviatesk: I tried to swap these colourings thinking about the consistency with ProfileView, but just found the profile view full of red colours is less cool haha:
image

pfitzseb: Hm, ideally we'd have a legend to explain what the colors mean (or maybe show "gc"/"dynamic dispatch" next to the function name).

aviatesk: I don't know which is better, but a legend will occupy some space in profiler view (which currently looks beautiful and cool) and will look a bit too much I guess ?

aviatesk: How about this ?
image

Just renaming classes and reuse them for the latter solution

pfitzseb: Looks good to me. Would be nice to have all text under the toolbar, if you want to take a look at that.

aviatesk: > Would be nice to have all text under the toolbar

sounds better, will push a commit later with the final image.

aviatesk: hm after taking some time to find a way to separate those texts into an new `span`, 
I found we need to tweak the `ToolBar` ~ `ProfileViewer` chain a bit and it's somewhat not just easy -- I think we need some additional logic to accept optional texts under the toolbar here:
https://github.com/JunoLab/atom-ink/blob/7803ad2ce517a4e985d945de24af7e81b6b7143d/lib/plots/pane.js#L60
but still not figured out how exactly I should do.

I can tackle this again when I have time (maybe weekend I guess, I'm currently rather busy for my work), but do you want me to push the changes that at least implements the above image for now ?

pfitzseb: Yeah, go for it. We can move the text around in a follow-up PR.

aviatesk: okay !

PR #674, opened 2020-01-30
highlight current cell

pfitzseb:
image

Fixes https://github.com/JunoLab/Juno.jl/issues/449.

### DIFF
--- a/lib/misc/cells.js
+++ b/lib/misc/cells.js
@@ -4,7 +4,7 @@ import { get as weaveGet,
          moveNext as weaveMoveNext,
          movePrev as weaveMovePrev } from './weave.js'
 
-function getRange (ed) {
+export function getRange (ed) {
   // Cell range is:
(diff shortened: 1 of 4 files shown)


aviatesk:
@pfitzseb This is nice enhancement, but highlighting code can appear weird in some cases because our cell detection strategy is not so matured yet:
e.g.:
```julia
"""
    foo(args)

docs

### example # recognized as cell

\`\`\`julia
things
\`\`\`
"""
function foo(args)
  body
end

## comments # recognized as cell
...
```

I haven't looked into cell detection strategy yet, but are there problems other than `##`s in docstring are recognized as cell ?
If it will take a time to improve cell detection strategy, we may be better to consider making a config setting to disable this.

pfitzseb:
That's just a bug then (which also applies to execution). Should be easy enough to fix the string case (and I don't think there should be any others).

That said, a setting still makes sense imo.

PR #677, opened 2020-02-04
add option for format-on-save

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/363.

### DIFF
--- a/lib/runtime/formatter.js
+++ b/lib/runtime/formatter.js
@@ -1,6 +1,7 @@
 /** @babel */
 
 import { client } from '../connection'
+import { CompositeDisposable } from 'atom'
 
 const format = client.import('format')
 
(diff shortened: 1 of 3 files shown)


lib/runtime/formatter.js
```diff
@@ -42,7 +43,7 @@ function formatEditorTextInRange (editor, range, text) {
         dismissable: true
       })
     } else {
-      if (marker.isValid()) {
+      if (marker.isValid() || force) {
```
aviatesk: When will those `force` options be used ?

lib/runtime/formatter.js
```diff
@@ -66,3 +67,35 @@ function markRange(editor, range) {
...
+          formatEditor(ed).then(() => {
+            edWatch.delete(ed)
+            ed.save().then(() => {
+              edWatch.add(ed)
+            })
+          })
+        }
+      }
+    })
+    subs.add(edsub)
```
aviatesk: I guess we can simplify the logic by using [`TextBuffer.onWillSave`](https://flight-manual.atom.io/api/v1.43.0/TextBuffer/#instance-onWillSave) (and removing `edWatch`) ?

lib/runtime/formatter.js
pfitzseb: Never, I just forgot to remove this from an earlier iteration.

lib/runtime/formatter.js
pfitzseb: No, because the order in which `onWillSave` hooks run isn't really specified (technically it is based on package activation order, I think). 
If we were to just use `onWillSave` then we'd grab a certain buffer state and invoke Julia's formatter, but in the meantime the whitespace package might modify the code, which means that our formatter fails (hence the `force` option I had above).

aviatesk:
sorry, wait a minute.

PR #678, opened 2020-02-05
option for ctrl/cmd modifier when clicking link

pfitzseb:
Requires https://github.com/JunoLab/atom-ink/pull/251.


aviatesk:
@pfitzseb is it possible/easy to suppress underline highlighting on link-hover when without ctrl/cmd modifier ?
Currently the highlight is invoked always on link-hover and could be misleading I guess.

PR #679, opened 2020-02-06
return 0 if there is no local context

aviatesk:
Fixes https://github.com/JunoLab/Atom.jl/issues/237


aviatesk:
I checked all the usages of `getLocalContext` and I'm so sure this won't break anything but just fix the rare case errors.
Going to merge this.

PR #682, opened 2020-02-06
allow formatting options

aviatesk:
Will work with: https://github.com/JunoLab/Atom.jl/pull/259

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -6,6 +6,7 @@ config =
     default: 'julia'
     description: 'The location of the Julia binary.'
     order: 1
+
   juliaOptions:
     type: 'object'
     order: 2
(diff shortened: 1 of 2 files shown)


aviatesk:
Rebase after https://github.com/JunoLab/atom-julia-client/pull/677 gets merged

lib/package/config.coffee
```diff
@@ -105,6 +106,75 @@ config =
...
         default: false
         order: 13
+      formattettingOptions:
+        title: 'Formatting options'
+        description: 'Specifies [formatting options for `JuliaFormatter.format_text`](https://domluna.github.io/JuliaFormatter.jl/stable/#Formatting-Options-1).'
+        type: 'object'
+        order: 14
+        collapsed: true
+        properties:
+          indent:
```
pfitzseb: I don't think we need that option -- it's intuitive to have JuliaFormatter follow the current editor, no?

lib/package/config.coffee
```diff
@@ -105,6 +106,75 @@ config =
...
+            title: 'indent'
+            description:
+              '''
+              Specifies the number of spaces used for an indentation.
+              With the default value (`-1`), `indent` will be automatically set to the current editor's [tab length](https://flight-manual.atom.io/using-atom/sections/basic-customization/#configuration-key-reference).
+              '''
+            type: 'integer'
+            default: -1
+            order: 1
+          margin:
```
pfitzseb: Same here, but maybe to a lesser extent.

lib/package/config.coffee
aviatesk: It's very intuitive that `indent` and `margin` follow the current editor's ones, but why I added them is that "some" project may want members to have the same setting of `indent` and `margin` for formatting within a project, in the future.
Those `indent` and `margin` may be different from those an user have specified as their general (or language specific) setting.
What I'm thinking about is somewhat will be enabled by https://github.com/domluna/JuliaFormatter.jl/issues/184.

***

But at the same time, I feel those settings are kinda too much for now -- I don't think currently there is a Julia project that encourages members to use the exactly same setting even for local formatting task.

So as a conclusion I'm totally okay to remove these configs for now.

lib/package/config.coffee
```diff
@@ -105,6 +106,75 @@ config =
         type: 'boolean'
         default: false
         order: 13
+      formattettingOptions:
```
waldyrious:
```suggestion
      formattingOptions:
```

aviatesk: thanks :)

lib/package/config.coffee
pfitzseb: Let's keep this for now then.

PR #683, opened 2020-02-10
debug file

pfitzseb:
Requires https://github.com/JunoLab/Atom.jl/pull/269.
Requires https://github.com/JunoLab/atom-ink/pull/255.
Requires https://github.com/JunoLab/atom-ink/pull/256.

### DIFF
--- a/keymaps/julia-client.cson.cmd
+++ b/keymaps/julia-client.cson.cmd
@@ -9,6 +9,8 @@ Any global commands should either be non-default or, ideally, prefixed with `C-J
 '.platform-darwin atom-text-editor[data-grammar="source julia"]:not(.mini),
   ink-terminal.julia-terminal,
   .ink-debugger-container':
+  'f5': 'julia-debug:run-file'
+  'cmd-f5': 'julia-debug:step-through-file'
   'shift-f5': 'julia-debug:stop-debugging'
   'f8': 'julia-debug:continue'
   'shift-f8': 'julia-debug:step-to-selected-line'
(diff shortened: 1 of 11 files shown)


pfitzseb:
Peek 2020-02-10 16-57

aviatesk:
Do you want
- tool-bar icon for invoking file debugging ?
- menus ?

pfitzseb:
Ah, right, forgot about that. We can now also add commands to debug the currently selected block etc.

aviatesk:
> We can now also add commands to debug the currently selected block

Maybe `Julia Debug: Run Block` and `Julia Debug: Step Through Block` ?

pfitzseb:
Yeah, that sounds goood.

aviatesk:
> > We can now also add commands to debug the currently selected block
> 
> Maybe `Julia Debug: Run Block` and `Julia Debug: Step Through Block` ?

Done ;) Also added `Run Cell` and `Step Through Cell`.

aviatesk:
@pfitzseb don't you mind renaming `run-and-move` to `run-block-and-move` ?

aviatesk:
As for menus, how about those submenus `Run` and `Debug` ?:
image

pfitzseb:
image

pfitzseb:
> @pfitzseb don't you mind renaming run-and-move to run-block-and-move ?

Mh, I'd like to keep commands that are likely used by others the same, but maybe it's fine to rename this one.

> As for menus, how about those submenus Run and Debug ?:

I don't like deeply nested menus much. Imho the current structure is still fine (and we don't have too many items yet).

aviatesk:
> Mh, I'd like to keep commands that are likely used by others the same, but maybe it's fine to rename this one.

Yeah, this may break some user's own keybinding. Suggested because you did some renaming around debugging commands, but yes `run-and-move` is so much often used that them I guess.

> I don't like deeply nested menus much. Imho the current structure is still fine (and we don't have too many items yet).

Okay, let's just keep menus then.

PR #685, opened 2020-02-17
fix block finding

pfitzseb:
Fixes both multiline backtick strings and modules:
````
```
kjgkg
```
module Foo
1+1
2+2
3+3
end
````

Regression tests would probably help a lot here, but  ¯\\\_(ツ)\_/¯

### DIFF
--- a/lib/misc/blocks.js
+++ b/lib/misc/blocks.js
@@ -19,12 +19,15 @@ function isBlank ({line, scope}, allowDocstrings = false) {
   return /^\s*(#.*)?$/.test(line)
 }
 function isEnd ({ line, scope }) {
-  for (const s of scope) {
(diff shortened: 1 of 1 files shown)


aviatesk:
`sp/fixblockfindingagainandagainandagain` omg 😅 

> Regression tests would probably help a lot here, but ¯\_(ツ)_/¯

I thought about the exact same thing yesterday actually.
I would like to open an issue about this on Juno.jl or this repository as a reminder, and thinking about possible contribution from others.

Also what do you think about using the Julia runtime (especially CSTParser I mean) here as well ?
I think about this only naively, but since most features using `blocks.js` rely on running Julia anyway, I guess it doesn't limit things but we can use our fav language for this as well (and then we can write tests easily as well).
We need to make `cells.js` not rely on `getLines` in `blocks.js` then though.

PR #686, opened 2020-02-17
rotate step-in/-out buttons

pfitzseb:
image

### DIFF
--- a/styles/julia-client.less
+++ b/styles/julia-client.less
@@ -163,3 +163,11 @@ atom-text-editor.editor[data-grammar$="source weave latex"] {
   border-top: solid fade(@syntax-text-color, 80%) 1px;
   margin-top: -1px;
 }
+
+.ink-debugger-container .ink-debug-toolbar .btn {
+  &.icon-alignment-align, &.icon-alignment-aligned-to {
+    &::before {
+      transform: translateY(3px) rotate(225deg) scale(0.8)
+    }
+  }
+}


ssfrr:
I'm really tweaking the shade of the bikeshed here, but would it be possible to tilt them so they're facing to the right? so "step-into" is down-right, and "step-out-of" is up-right?

pfitzseb:
Shouldn't step out point to the upper left then?

ssfrr:
My mental model is that time goes left-to-right. And time marches inexorably onwards. More specifically, when you step-out of a function you don't end up returning to where you came from, you've progressed later into the upper stack frame

DNF2:
Hmm. I'm always thinking, 'down' is the next line of code, so the arrow of time is downwards.

But there's probably no good way to conform to everyones mental model. The way people visualize time, for example, varies a lot. Some see the year moving 'circularly counter clockwise', some clockwise, some along a square perimeter, some like a three-dimensional askew torus, etc. etc. And some are _really_ weird.

ssfrr:
Sure that's fair. I'm happy with whatever outcome.

Also @jkrumbiegel mentioned he might take a crack at something as well.

aviatesk:
I'm afraid but I would like to tell my understanding -- at first "down" to step out and "up" to step into looked very weird to me, but then I could convince myself that they make sense in terms of they are very consistent with how a callstack and workspace view appear.
So for now, at least to me, making them vertical while keeping them "down" to step out and "up" to step into might be best.
Well, I really agree with that not everyone will respect consistency with other features here ...

jkrumbiegel:
I posted this in the slack channel as well, could be another idea:

The blue line shows the "until selected line" concept a bit better maybe, and for me the step in and out versions are a bit more intuitive.

grafik

ssfrr:
Lovely - Do you think it would be good to shorten the arrow on the "until selected line" icon so it stops at the line? Might be visually more noisy so maybe not a good idea.

jkrumbiegel:
I thought the same, but then again that might be taken to mean that the other arrow goes over multiple lines, although it's just supposed to evoke "down" and "lines"

jkrumbiegel:
grafik

jkrumbiegel:
how about this instead? 
grafik

ssfrr:
gorgeous.

pfitzseb:
Closing in favor of https://github.com/JunoLab/atom-julia-client/pull/693.

PR #695, opened 2020-02-24
improve interrupt handling

pfitzseb:
This completely gets rid of any special casing on Windows and instead pipes all interrupts through the REPL (which is now possible because we run a blocking task in the REPL backend during inline evaluation). 

In theory this will tremendously improve interrupts on Windows, work around the fact that the `spawnInterruptible` script is detected as a virus by some antivir software, and delete a whole bunch of code. ~~In practice this isn't quite there yet because we still have a bunch of race conditions to take care of -- I _think_ we need to use only one backend task (like the REPL, basically) and push eval work into that via a `Channel`.~~

Requires https://github.com/JunoLab/Atom.jl/pull/291.


aviatesk:
Can give a review and test later this week :)

pfitzseb:
Ok, this should be in pretty good shape now.

aviatesk:
Fancy. I gonna have a time tmrw to review on those recent REPL woes, but I'm okay if you merge this if you want.

pfitzseb:
I think I'd rather wait for you to take a look at this considering how messy the releases last week were :P

PR #699, opened 2020-03-04
Fix Atom crashes on Windows

pfitzseb:
also a bit of code clean up.

### DIFF
--- a/lib/runtime/console.js
+++ b/lib/runtime/console.js
@@ -336,7 +336,7 @@ function shellPty (cwd) {
           cwd = paths.home()
         }
         const env = customEnv()
-        const pty = pty.spawn(atom.config.get("julia-client.consoleOptions.shell"), [], {
+        const ty = pty.spawn(atom.config.get("julia-client.consoleOptions.shell"), [], {
           cols: 100,
           rows: 30,
           cwd: cwd,
@@ -345,7 +345,7 @@ function shellPty (cwd) {
           handleFlowControl: true
         })
         resolve({
-          pty: pty,
+          pty: ty,
           cwd: cwd})
       } else {
         reject()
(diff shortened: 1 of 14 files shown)


lib/runtime/modules.coffee
```diff
@@ -84,8 +86,12 @@ module.exports =
       @updateForEditor item
     else
       mod = item.juliaModule or 'Main'
-      ismodule(mod).then (ismod) =>
-        @setCurrent main: mod, inactive: !ismod
+      m = ismodule(mod)
+      m
```
aviatesk:
```suggestion
      ismodule(mod)
```

lib/runtime/modules.coffee
```diff
@@ -64,6 +64,8 @@ module.exports =
           else if @lastEditorModule?
             modules.unshift @follow
           modules
+        module.catch (err) =>
```
aviatesk:
```suggestion
        modules.catch (err) =>
```

lib/runtime/modules.coffee
```diff
@@ -105,7 +111,8 @@ module.exports =
       code: ed.getText()
       row: row+1, column: column+1
       module: ed.juliaModule
-    getmodule(data)
+    getmodule(data).catch (err) =>
+      console.log err
```
aviatesk: nitpick, but would be more consistent with the other parts
```suggestion
    getmodule(data)
      .catch (err) =>
        console.log err
```

lib/runtime/console.js
```diff
@@ -334,14 +336,16 @@ function shellPty (cwd) {
           cwd = paths.home()
         }
         const env = customEnv()
+        const pty = pty.spawn(atom.config.get("julia-client.consoleOptions.shell"), [], {
```
aviatesk: This line will cause undef property error since the global name `pty` is shadowed the local variable.
Need to fix like below:
```suggestion
        const p = pty.spawn(atom.config.get("julia-client.consoleOptions.shell"), [], {
```
and also rename the `pty` in `resolve`

aminya:
Thanks for fixing this! ❤️

PR #704, opened 2020-03-15
Type script

aminya:
Converting codebase to TypeScript
- [X] adding configurations
- [X] decaffeinate
- [X] js-to-ts
- [X] linting for JS, TS, Coffee
- [X] Use Github actions instead of Travis and Appveyor. See the running actions here: https://github.com/aminya/atom-julia-client/actions?query=branch%3ATypeScript
- [ ] Compiling each ts file one by one to replace the corresponding `lib` file
		- Using import/exports instead of requiring if possible
		- Adding type definitions
		- other ts fixes
		- tslint
		- Using [atom-select-list-ts](https://github.com/aminya/atom-select-list/tree/TypeScript) instead of space-pen-views using the similar that was made for [atom-indent-detective](https://github.com/JunoLab/atom-indent-detective/pull/17)
		- Using [etch-ts](https://github.com/atom/etch/pull/82) instead of etch
		- Optimizing code wherever possible
		- Using TypeAssembly if worth it

	- [ ] connection\process\basic.ts
	- [ ] connection\process\boot.ts
	- [ ] connection\process\cycler.ts
	- [ ] connection\process\remote.ts
	- [ ] connection\process\server.ts
	- [ ] connection\process\tcp.ts
	- [ ] connection\client.ts
	- [ ] connection\ipc.ts
	- [ ] connection\local.ts
	- [ ] connection\messages.ts
	- [ ] connection\terminal.ts
	- [x] misc\blocks.ts
	- [x] misc\cells.ts
	- [x] misc\colors.ts
	- [ ] misc\paths.ts
	- [x] misc\scopes.ts
	- [ ] misc\weave.ts
	- [ ] misc\words.ts
	- [ ] package\commands.ts
	- [ ] package\config.ts
	- [ ] package\menu.ts
	- [ ] package\settings.ts
	- [ ] package\toolbar.ts
	- [ ] runtime\completions.ts
	- [ ] runtime\console.ts
	- [ ] runtime\datatip.ts
	- [ ] runtime\debugger.ts
	- [ ] runtime\debuginfo.ts
	- [ ] runtime\evaluation.ts
	- [ ] runtime\formatter.ts
	- [ ] runtime\frontend.ts
	- [ ] runtime\goto.ts
	- [ ] runtime\linter.ts
	- [ ] runtime\modules.ts
	- [ ] runtime\outline.ts
	- [ ] runtime\packages.ts
	- [ ] runtime\plots.ts
	- [ ] runtime\profiler.ts
[...]

PR #705, opened 2020-03-19
Typescript init

aminya:
## Ready to get merged
Please merge this so I can continue with other PRs

## **This PR doesn't change any behavior of Julia Client.** 

# Initialization for Converting codebase to TypeScript: 
This is the initialization of that PR. 

I am distilling https://github.com/JunoLab/atom-julia-client/pull/704 into smaller PRs, so I can track the changes easier, and making reviewing the actual editing part easier for you.

All the code is stored in (lib_src) and doesn't change any source which is in use.

# Config
- [X] adding configurations for future conversion
- [X] linting for TS, JS, Coffee (ESLint + TSLint)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,80 @@
+{
+  // JS Files
+  "root": true,
+  "env": {
+    "atomtest": true,
+    "es6": true,
+    "node": true,
+    "browser": true,
+    "jasmine": true
+  },
+  "globals": { "atom": "writable" },
+  "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaFeatures": { "jsx": true },
+    "ecmaVersion": 2018,
+    "sourceType": "module"
+  },
+  "plugins": ["only-warn"],
+  "extends": ["eslint:recommended"],
+  "overrides": [
+    { // Bundled node version with atom has an old ESLint
+      // TypeScript files
+      "files": ["**/*.ts", "**/*.tsx"],
+      "env": {
+        "atomtest": true,
+        "es6": true,
+        "node": true,
+        "browser": true
+      },
+      "globals": { "atom": "writable" },
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "ecmaFeatures": { "jsx": true },
+        "ecmaVersion": 2018,
+        "sourceType": "module"
+      },
+      "plugins": ["@typescript-eslint", "only-warn"],
+      "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended"
+      ],
+      "rules": {
+        "@typescript-eslint/explicit-function-return-type": "off",
+        "@typescript-eslint/camelcase": "off",
+        "@typescript-eslint/no-use-before-define": "off",
+        "@typescript-eslint/member-delimiter-style": "off"
+      }
+    },
+    {
+      // CoffeeScript files
+      "files": ["**/*.coffee"],
+      "env": {
+        "atomtest": true,
+        "es6": true,
+        "node": true,
+        "browser": true
+      },
+      "globals": { "atom": "writable" },
+      // "parser": "eslint-plugin-coffee",
+      "parser": "eslint-plugin-coffee",
+      "parserOptions": {
+        "ecmaFeatures": { "jsx": true },
+        "ecmaVersion": 2018,
+        "sourceType": "module"
+      },
+      "plugins": ["coffee", "only-warn"],
+      "extends": ["plugin:coffee/eslint-recommended"]
+    },
+    {
+      // JSON files
+      "files": ["*.json"],
+      "plugins": ["json"],
+      "extends": ["plugin:json/recommended"],
+      "rules": {
+        "json/*": ["error", {"allowComments": true}]
+      }
+    }
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .DS_Store
 npm-debug.log
 node_modules
+lib_src/**/*
+!lib_src/tsconfig.json
+!lib_src/tslint.json
+tsconfig.tsbuildinfo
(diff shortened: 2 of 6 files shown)


aminya:
@pfitzseb @aviatesk Could you please merge this PR?

aminya:
@pfitzseb @aviatesk Could you please merge this PR?

aminya:
I even simplified this PR more. I will do the whole process file by file in separate PRs, so you guys can continue developing without any conflicts.

So please merge this PR as soon as possible. This is blocking me from continuing.

aviatesk:
Hi @aminya, sorry for taking a time to get back to you.

We would very much welcome JSfying of our code base (in per-file basis), but I really don't appreciate conversions to TypeScript.

The main reason is that we're planning the migration of ink and julia-client in the coming months and it will be such a big refactoring.
At this point we don't want to give the other huge change to our code base but rather keep our code base as it has been, in order to perform the migration safely.

TS _seems_ more promising for current front-end development in general, and it would be fancy to have type-safety and more extensive development utilities for Juno, _in the future_.
But this is clearly not our priority for now.
And to say more, I'm not sure TS will really help our code base, at least as far as we live with Atom:
- type-safety provided by TS isn't strict/true -- the case is much worse especially when we have lots of dependencies that don't provide (reliable) type definitions (like etch for our case)
- Atom doesn't support native TS compiler, and so TS will add another complexity for our development process
- lots of development utilities provided by TS-server are available for JS files

aminya:
> 
> 
> Hi @aminya, sorry for taking a time to get back to you.
> 

Hi! No problem. 

> We would very much welcome JSfying of our code base (in per-file basis), but I really don't appreciate conversions to TypeScript.
> 
> The main reason is that we're planning the migration of ink and julia-client in the coming months and it will be such a big refactoring.
> At this point we don't want to give the other huge change to our code base but rather keep our code base as it has been, in order to perform the migration safely.
> TS _seems_ more promising for current front-end development in general, and it would be fancy to have type-safety and more extensive development utilities for Juno, _in the future_.
> But this is clearly not our priority for now.


Regarding this ink and julia-client merging, I have talked to @pfitzseb, and we said that it worth it to convert in the merging process. I would be happy to help merging Ink, but I think julia-client itself can be enhaced on its own without mering Ink.


> And to say more, I'm not sure TS will really help our code base, at least as far as we live with Atom:
> 
>     * type-safety provided by TS isn't strict/true -- the case is much worse especially when we have lots of dependencies that don't provide (reliable) type definitions (like etch for our case)
> 
>     * Atom doesn't support native TS compiler, and so TS will add another complexity for our development process
> 
>     * lots of development utilities provided by TS-server are available for JS files


I already talked about the many benefits here: https://github.com/JunoLab/atom-indent-detective/pull/17#issuecomment-599954980, but I didn't get any response from you.

> 
> - You don't need to do anything regarding TS. This package is good to go and will act like a JS package for usage (JS is already built).
> 
> - Typescript isn't about just tracking the implementation. It brings many improvements over simple JavaScript (enhances JS without drawbacks):
> 	- it reveals many things that are not visible when you use JS. Things like type-changing of the variables, if a function returns the type we want, performance improvement by using better types, etc. You see this package loads 1/4 of the original, and its runtime was better in my measurements (30-50%).
> 	- IntelliSense/linting
> 	- API documentation on the fly. It allows others to understand what the code is doing.
> 	- You can write your normal JS code in TypeScript and it works. TypeScript infers many of the types automatically. 
> 	- It removes the need for babel or any other processor.
> 	- We can compile TypeScript to Wasm using AssemblyScript which opens a whole new world for improving the package. For example, I plan to compile the `getIndet` function of this package.
> 	- The future isn't plain JavaScript. For example, VSCode is already using TS (which IMO has better performance while I like atom more).
> 
> Read this article to know more about the advantages:
> https://dzone.com/articles/what-is-typescript-and-why-use-it


- TypeScript is not anything different from JavaScript. 
- It is just a simple tool to modernize the code and see its flaws. 
- There is no learning curve. 
- It doesn't break anything. 
- It works just like JavaScript (no interop issues or anything).

- Just look at the commits history and you see a lot of optimization, TODOs, flaws I have found:
https://github.com/JunoLab/atom-julia-client/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc


Last point:
**Since I am doing the work for you (in a systematic way that is easy to follow), I don't really understand the reason for the objection and resistance.**

aviatesk:
I'm afraid to say, but did you _really_ read my opinions ? What do you think on them ?

> type-safety provided by TS isn't strict/true -- the case is much worse especially when we have lots of dependencies that don't provide (reliable) type definitions (like etch for our case)
> Atom doesn't support native TS compiler, and so TS will add another complexity for our development process
> lots of development utilities provided by TS-server are available for JS files

You should always try to see pros and cons of things. And since this change is really big one, I would like to figure out it is really necessary to be done.

aminya:
> > * type-safety provided by TS isn't strict/true -- the case is much worse especially when we have lots of dependencies that don't provide (reliable) type definitions (like etch for our case)
> > * lots of development utilities provided by TS-server are available for JS files

I mentioned the answers to these in my points: TypeScript is just JavaScript. lack of type definitions doesn't break the code or doesn't mean that it is not compiled. It just means that the code will be copied (compiled) like it is a simple JavaScript. 

The reason that you think this adds complexity is that you think I am changing JavaScript to C++ for example 😄 while I am just fixing the code issues itself.

> > * Atom doesn't support native TS compiler, and so TS will add another complexity for our development process

There is a compiler: https://github.com/smhxx/atom-ts-transpiler. But we don't wanna rely on the user's computer for compiling anyways. It makes sense to deliver ready JS files to the user instead.

aminya:
I separate Github Actions part of this PR into another PR https://github.com/JunoLab/atom-julia-client/pull/712, the failiur of the linter on Travis is because of this

aviatesk:
So after all, I would like to decline you suggestions for migrating into TS. Reasons:
- the additional complexity of transpiling, declaring types, etc will be very annoying for this package, at least at this stage when the code base is still under rapid development
- we can't get type-safety after all; we have dependencies that don't provide (valid) type definitions, and the part with interaction with ink is all untyped

PR #711, opened 2020-03-30
allow julia path relative to atom executable

pfitzseb:


lib/misc/paths.coffee
```diff
@@ -17,17 +17,30 @@ module.exports =
...
+      current_dir = process.cwd()
+      exepath = path.dirname(process.execPath)
+      
+      try
+        process.chdir(exepath)
+        realpath = fs.realpathSync(p)
+        if fs.existsSync(realpath) then resolve(realpath)
+      catch err
+        try
+          process.chdir(current_dir)
```
aviatesk: Why isn't this in `finally` block ?

pfitzseb: It is now ;) Good catch.

pfitzseb:
That's useful to have for portable installations and worked up until Juno 0.11 because of the powershell wrapper (and it got requested by @RoyiAvital).

aviatesk:
makes sense 😊👍

jkaet:
Looks like this issue returned with decaffination?

pfitzseb:
I don't think so. We just never made a release that included this fix.

PR #718, opened 2020-04-08
disable process cycler by default

pfitzseb:


aviatesk:
https://github.com/JunoLab/atom-julia-client/blob/6cfcca10b097adf5a3d3535d993803b3df965740/lib/package/config.coffee#L19
I guess this should be "two" instead of "three", no ?

PR #720, opened 2020-04-10
just insert one line when we reached to the end of buffer

aviatesk:


kescobo:
Is it possible to make this a user-settable parameter? I prefer the current behavior, but can understand wanting this one.

PR #721, opened 2020-04-10
Better default for cell delimiters

jonathan-laurent:
Some style guides (such as [YASGuide](https://github.com/jrevels/YASGuide)) recommend using the following section headers:

```
#####
##### Section Name
#####
```

This conflicts with the default cell delimiters regexps of Juno, which would create three cells when seeing the header above.

**Proposed fix:** replace the `##` regexp by `##\s`, as suggested by @pfitzseb on Slack.

Signed-off-by: Jonathan Laurent <jonathan.laurent@cs.cmu.edu>

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -311,7 +311,7 @@ config =
       cellDelimiter:
         title: 'Cell Delimiter'
         type: 'array'
-        default: ['##', '#---', '#%%', '# %%']
+        default: ['##\\s', '#---', '#%%', '# %%']
         description: 'Regular expressions for determining cell delimiters.'
         order: 11
       highlightCells:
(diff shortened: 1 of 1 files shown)


lib/package/config.coffee
```diff
@@ -311,7 +311,7 @@ config =
       cellDelimiter:
         title: 'Cell Delimiter'
         type: 'array'
-        default: ['##', '#---', '#%%', '# %%']
+        default: ['##\\s', '#---', '#%%', '# %%']
```
aviatesk, line 314:
```suggestion
        default: ['##\\s', '#---', '#\\s?%%']
```

aviatesk:
thanks @jonathan-laurent !

Issue #715, opened 2020-04-01
Decaffination + Modernize import/export 


aminya:
# Method
Decaffeination can be done easily using:
```
decaffeinate . --use-js-modules --loose
```

or without automatic import/export
```
decaffeinate . --loose
```

We need to run either of those and then fix the linkage of modules manually. We don't need to use `import/export` for internal package deps, but it will help for external deps.

# Why
- Webpack build fails because of coffee files https://github.com/JunoLab/atom-julia-client/pull/714.

# Side-Note
P.S: There is also bulk-decaffeinate that automatically searches for connections, but it doesn't work well on Windows (I couldn't get it to work)
https://github.com/decaffeinate/bulk-decaffeinate
https://github.com/decaffeinate/bulk-decaffeinate/issues/265

aminya:
cc: @pfitzseb @aviatesk Let me know if you agree with this

To do this: 
- we should merge the open PRs (mainly  https://github.com/JunoLab/atom-julia-client/pull/641 and https://github.com/JunoLab/atom-julia-client/pull/713 that edit coffee files) 
- and stop editing the files for a few days (maybe a week)

This doesn't take much. We just need to fix the import/export connections for now. Decaffeinated already compiles a very good code, and fine adjustments can be done later.

aviatesk:
Decaffeination is very welcome. The most important thing is whether the JSfied code is maintainable; don't forget to read through and check the code "makes sense" even when you relied on the auto-decaf tool at the beginning.

As for webpack, I guess it only gives negligible loading time improvement. Atom uses transpiled cache if we don't transpile AOT. So it's not good idea to rush decaf stuff for webpacking, etc.

pfitzseb:
Yeah, let's do the decaffeination as a first step. Just a quick comment on your PRs, @aminya: We're currently using something close to [standard](https://standardjs.com/) for our JS, so it'd be cool if you could switch your linter pass to that (or just follow the style manually, I guess).

aminya:
Probably, I will not spend more time on this project. I cannot deal with loads of strong opinions.

If you want you can close this issue.

PR #723, opened 2020-04-13
JSfy paths.coffee

aviatesk:
Appreciate any kind of testing against this.


aminya:
Are we doing decaffeination? I didn't get any answer here: https://github.com/JunoLab/atom-julia-client/issues/715 or in Slack group.

PR #724, opened 2020-04-13
release note

aviatesk:
closes https://github.com/JunoLab/Juno.jl/issues/545
needs https://github.com/JunoLab/atom-ink/pull/270

workflow:
- upload images somewhere (GitHub, Discourse etc) -> write markdown -> publish it here and on discourse

when release note pop up:
- when an user update this package from the previous minor version => show the latest minor version release note
- when an user update from the current minor verison => show the latest patch version release note

### DIFF
--- a/lib/julia-client.coffee
+++ b/lib/julia-client.coffee
@@ -4,11 +4,14 @@ commands = require './package/commands'
 config = require './package/config'
 menu = require './package/menu'
 settings = require './package/settings'
+release = require './package/release-note'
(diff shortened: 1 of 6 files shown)


pfitzseb:
Very cool.
I think I'd prefer to have the pacth notes in a modal instead of a normal pane though.
Will look over the code later today or tomorrow.

aviatesk:
> have the pacth notes in a modal instead of a normal pane

I'm okay with this. Then we don't need to interact with ink.

PR #725, opened 2020-04-13
release note

aviatesk:
try to show release notes in a modal.

Couple of problems I couldn't solve:
- can't scroll on the modal
- can't close the modal on unfocus/<kbd>esc</kbd> event
- modal doesn't persist on startup
...

Please modify this PR if you figure out the fixes, or even supercede this if you want to start from the other way.

### DIFF
--- a/lib/package/release-note.js
+++ b/lib/package/release-note.js
@@ -13,22 +13,35 @@ const RELEASE_NOTE_DIR = path.join(__dirname, '..', '..', 'release-notes')
 export function activate (startupNoteVersion) {
   subs = new CompositeDisposable()
   const view = document.createElement('div')
-  view.classList.add('julia-client-release-note')
   const panel = atom.workspace.addModalPanel({ item: view })
-  const showNote = (version) => {
+
+  const inner = document.createElement('div')
+  inner.classList.add('julia-client-release-note')
+  inner.setAttribute('tabindex', "1")
+
+  const closeButton = document.createElement('button')
+  closeButton.classList.add('btn', 'icon', 'icon-remove-close', 'release-note-close-button')
+
+  view.appendChild(closeButton)
+  view.appendChild(inner)
+
+  closeButton.onclick = () => panel.hide()
+
+  const showNote = (version, versions) => {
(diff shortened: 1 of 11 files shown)


aviatesk:
switched the base branch

pfitzseb:
Simplified the logic a bit and fixed the styling. Let me know what you think.

aviatesk:
> Simplified the logic a bit and fixed the styling. Let me know what you think.

nice, but... I'm afraid to say I'm now slightly against the modal idea; maybe some users want to see/try new features while having editor pane and release note pane, side by side ?
And I guess this is why vscode shows release note as an editor pane, I dunno.
Well, the modal surely has more attentions.

lib/package/release-note.js
```diff
@@ -0,0 +1,64 @@
...
+        .then(version => showNote(version.replace(/v\s(.+)/, '$1')))
+        .catch(err => console.log(err))
+    }),
+    atom.commands.add(panelView, 'julia-client:cancel-release-note',  (event) => {
+      close()
+      event.stopPropagation()
+    }),
+    atom.packages.onDidActivateInitialPackages(() => {
+      if (startupNoteVersion) setTimeout(() => showNote(startupNoteVersion), 500)
+    }),
```
aviatesk, line 55: on my setup, subscribing on `onDidActivateInitialPackages` was necessary for the modal to persist

pfitzseb: "persist" as in "not close directly after showing it"? Kinda makes sense, I suppose.

aviatesk: I tried to mean, the modal doesn't keep to be shown up without `onDidActivateInitialPackage` maybe because otherwise some other modal implicitly show up on startup and hide (already shown) the release note modal.
I don't think this fix is so robust though

pfitzseb:
> nice, but... I'm afraid to say I'm now slightly against the modal idea; maybe some users want to see/try new features while having editor pane and release note pane, side by side ?

I tend to agree :)

aviatesk:
okay I reverted back to the previous idea. Now this PR needs https://github.com/JunoLab/atom-ink/pull/270.

PR #731, opened 2020-04-15
Handling Windows paths with spaces in the REPL

FAlobaid:
Fixes JunoLab/Juno.jl#421 to correctly handle Windows paths with spaces in Juno REPL. Related pull request: JunoLab/Atom.jl#320.


lib/runtime/console.js
```diff
@@ -17,7 +17,7 @@ const { changeprompt, changemodule, fullpath } =
 
 const isWindows = process.platform === 'win32'
 const uriRegex = isWindows ?
-  /(@ ([^\s]+)\s(.*?)\:(\d+)|((([a-zA-Z]:|\.\.?|\~)|([^\0<>\?\|\/\s!$`&*()\[\]+'":;])+)?((\\|\/)([^\0<>\?\|\/\s!$`&*()\[\]+'":;])+)+)(\:\d+)?)/ :
+  /(@ ([^\s]+)\s(.*?)\:(\d+)|((([a-zA-Z]:|\.\.?|\~)|([^\0<>\?\|\/\s!$`&*()\[\]+'":;])+)?((\\|\/)([^\0<>\?\|\/!$`&*()\[\]+'":;])+)+\.[^\s:]+)(\:\d+)?)/ :
```
aviatesk, line 20: um then `\.[^\s:]+` pattern extends to "non-path" characters, e.g. when you have `"C:\\Users\\avi\\julia\\JunoLab\\hoge hoge.jl"` (note quotes in both ends), the regex match extends the final `"` and it makes an opener fails at file existence check.

FAlobaid: Thanks and sorry about that. I tried to do a minimal change fixing the problem 😊. Now, I'll replace `\.[^\s:]+` with `\.[^\0<>\?\|\/\s!$`&*()\[\]+'":;]+` which should correct it.

I'll update the pull request and the other one as well JunoLab/Atom.jl#320.
Thanks.

aviatesk: yeah, it would work, I think.
never to say sorry, we're very happy to have your contribution !

pfitzseb:
Yeah, I think this is strictly better than what we had before. Still not working perfectly when there's no extension, but whatever. Let's make the same change on unixes after merging this PR.

PR #734, opened 2020-04-16
Improved handling of paths with spaces in the REPL  (Win, macOS, Linux)

FAlobaid:
This fix is related to JunoLab/Juno.jl#421 and improves on the previous PR JunoLab/atom-julia-client#731 to better handle paths with spaces in the REPL for Windows, macOS and Linux. Among other things, it correctly handles extension-less files and newline. However, it doesn't fully match the following (so avoid):

1- If you start with a folder name with spaces (obviously). See example 1 in the avoid section.
2- If you have extension-less filename with spaces. See examples 2 and 3 in the avoid section.
3- If you have file extension with spaces. See examples 4 and 5 in the avoid section.

Related PR: JunoLab/Atom.jl#322

Examples for Windows:
win_js

Examples for macOS / Linux:
nix_js


lib/runtime/console.js
```diff
@@ -17,8 +17,8 @@ const { changeprompt, changemodule, fullpath } =
 
 const isWindows = process.platform === 'win32'
 const uriRegex = isWindows ?
-  /(@ ([^\s]+)\s(.*?)\:(\d+)|((([a-zA-Z]:|\.\.?|\~)|([^\0<>\?\|\/\s!$`&*()\[\]+'":;])+)?((\\|\/)([^\0<>\?\|\/!$`&*()\[\]+'":;])+)+\.[^\0<>\?\|\/\s!$`&*()\[\]+'":;]+)(\:\d+)?)/ :
-  /(@ ([^\s]+)\s(.*?)\:(\d+)|(((\.\.?|\~)|([^\0\s!$`&*()\[\]+'":;\\])+)?(\/([^\0\s!$`&*()\[\]+'":;\\])+)+)(\:\d+)?)/
+  /(@ ([^\s]+)\s(.*?)\:(\d+)|((([a-zA-Z]:|\.\.?|\~)|([^\0<>\?\|\/\s!$`&*()\[\]+'\":;])+)?((\\|\/)([^\0<>\?\|\/\n\r!$`&*()\[\]+'\":;])+)+(\.|\\|\/)[^\0<>\?\|\/\s!$`&*()\[\]+'\":;]+)(\:(\d+))?)/ :
```
aviatesk, line 20: why do we need `\` before `"` only on Windows ?

FAlobaid: Your're right these are not necessary here as they are artifact from me testing Julia first. It works either way, but I'll remove them for consistency.

Thanks.

FAlobaid:
> thanks for your help and sorry for late review

Most welcome and thanks for all your efforts :+1:

PR #735, opened 2020-04-20
support .JuliaFormatter.toml config file

aviatesk:
corresponds to: https://github.com/domluna/JuliaFormatter.jl/pull/221


pfitzseb:
Feels good to not have to support all of those settings anymore :)

aviatesk:
hehe, glad to hear from you ;)

PR #736, opened 2020-04-24
recover OS notification

aviatesk:
OS notifications is still an issue ? 
Searching in atom repo, there seems to be a problem in tree-view stuff in electron<4 on macOS, but couldn't find an issue related to OS notification.

### DIFF
--- a/lib/package/config.coffee
+++ b/lib/package/config.coffee
@@ -152,12 +152,12 @@ config =
           {value:'block', description:'Display results under code'}
           {value:'console', description:'Display results in the REPL'}
         ]
-        order: 1
+        order: 0
       scrollToResult:
         title: 'Scroll to Inline Results'
         type: 'boolean'
         default: false
-        order: 2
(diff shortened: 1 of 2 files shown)


pfitzseb:
https://github.com/electron/electron/issues/18701 is the upstream issue -- seems like it's only been fixed very recently and I don't think the new Electron version has propagated to Atom yet.

aviatesk:
okay.
Well, do you think we can disable `close` only on macOS ?
I found notifications are swept away after few seconds anyway, so I guess we actually don't need `close` calls from the beginning, no ?

I started to use Juno for tasks including time consuming computation, and I found I really want this OS notification.

(P.S. I also really can't reproduce the crash on my macOS Mojave 10.14.6 even when calling `close`s...)

pfitzseb:
It's probably fine to re-enable this fi we can test it on all platforms and keep it behind a setting.

PR #741, opened 2020-05-05
project monitor tile

aviatesk:
image

requires https://github.com/JunoLab/Atom.jl/pull/330

- [x] basic implementaion
- [x] command export

### DIFF
--- a/lib/connection/client.coffee
+++ b/lib/connection/client.coffee
@@ -155,7 +155,7 @@ module.exports =
       @ipc.reset()
 
   clargs: ->
-    {precompiled, optimisationLevel, deprecationWarnings} =
+    {optimisationLevel, deprecationWarnings} =
       atom.config.get 'julia-client.juliaOptions'
     as = []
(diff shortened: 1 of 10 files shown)


aviatesk:
I feel like I want to unify the terminologies of "environment" and "project", currently we use:
- "environment" in frontend (commands, etc)
- "project" in backend (Julia side)

pfitzseb:
Maybe we should have `Env: JunoLab` in the tile and `Module: Atom` in the module tile to make it easier to figure out what one is looking at?

aviatesk:
sounds nice idea, it's really confusing when I had two subsequent `Atom` tiles.

aviatesk:
now they look:
image

I have to admit they now occupy even more spaces though, but who complains.

pfitzseb:
`E: JunoLab`/`M: Atom` maybe? Should be clear enough with a tooltip.

aviatesk:
> `E: JunoLab`/`M: Atom` maybe? Should be clear enough with a tooltip.

hm, imho I prefer the first idea (`Env: ` and `Module: `) ...

